PR #35, opened 2025-10-14
Add title and subtitle props to PieChart component

kevinmarr:
Add title and subtitle props to PieChart component


kevinmarr:
Screenshot 2025-10-13 at 9 12 42 PM

Spacing with the value label there is kinda jank, not sure how to fix that.

PR #69, opened 2026-01-31
Add agg() passthrough function for stylistic aggregate wrapping

kevinmarr:
Implements a new GSQL function `agg()` that acts as a passthrough wrapper for aggregate expressions. It validates that its argument is an aggregate and returns it unchanged, producing identical SQL output.

## Features
- Only accepts aggregate expressions (errors on non-aggregates)
- True passthrough - compiles to identical SQL
- Works with all aggregate types (built-in, computed, functions, percentiles)
- Comprehensive test coverage (17 test cases)
- Zero runtime overhead

## Examples
```gsql
-- Valid uses
select agg(sum(revenue)) from orders
select agg(total_revenue) from orders  -- where total_revenue is aggregative
select agg(count_if(amount > 30)) from orders
select agg(p50(distance)) from flights

-- Invalid uses (error: agg() can only wrap aggregate expressions)
select agg(revenue) from orders       -- revenue is not an aggregate
select agg(1+1) from orders            -- scalar expression
```

## Implementation
Added special-case handling in `analyzeFunctionCall()` that:
1. Validates exactly 1 argument
2. Analyzes the argument expression
3. Checks if `isAgg` flag is true
4. Returns the argument directly (true passthrough)

## Testing
- All 119 existing tests pass
- Added 17 new test cases covering all scenarios
- Manual verification in examples/flights confirms correct behavior

---

**Continue Tasks:** ▶️ 1 queued — [View all](https://hub.continue.dev/inbox?pr=https%3A%2F%2Fgithub.com%2Fgraphene-data%2Fgraphene%2Fpull%2F69&utm_source=github_pr&utm_medium=pr_body&utm_campaign=continue_tasks)

### DIFF
--- a/lang/functions.ts
+++ b/lang/functions.ts
@@ -16,6 +16,22 @@ export function analyzeFunctionCall (expr: SyntaxNode, scope: Scope): Expression
   let rawName = txt(expr.getChild('Identifier')).toLowerCase()
   let argNodes = expr.getChildren('Expression')
 
+  // Special handling for agg() - passthrough identity function
+  if (rawName === 'agg') {
+    if (argNodes.length !== 1) {
+      return diag(expr, 'agg() requires exactly one argument', errExpr)
+    }
+    
+    let argExpr = analyzeExpression(argNodes[0], scope)
+    
+    if (!argExpr.isAgg) {
+      return diag(argNodes[0], 'agg() can only wrap aggregate expressions', errExpr)
+    }
+    
+    // Return the argument directly - true passthrough
+    return argExpr
+  }
+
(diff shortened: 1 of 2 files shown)


rogueg:
Doesn't `agg` only make sense when referring to a column that is an agg measure? In all other cases, seems like it should be an error.

I'm curious how your testing went. Does this actually help?

kevinmarr:
Yeah, maybe we can let this cook a bit longer while I test it more. I realized I was using Opus instead of Sonnet so it could be that that was what was improving the outputs and not this agg() change...

PR #78, opened 2026-02-09
Drop malloy

rogueg:
Removes malloy as a dependency, and instead generates dialect-specific sql as part of our analysis process.

The most notable feature we've lost is symmetric aggregates, which Malloy supports but we've decided isn't that important and we're not going to support.

---

**Continue Tasks:** ▶️ 1 queued — [View all](https://hub.continue.dev/inbox?pr=https%3A%2F%2Fgithub.com%2Fgraphene-data%2Fgraphene%2Fpull%2F78&utm_source=github_pr&utm_medium=pr_body&utm_campaign=continue_tasks)


kevinmarr:
@rogueg Remind me - how does this behave now in the fan-out case? Allow (like regular SQL), or error?

kevinmarr (approved):

PR #103, opened 2026-02-18
Drop `expr as alias` syntax for tables

rogueg:
This lets us simplify the grammar, and we had dropped it from our docs anyway.

---

**Continue Tasks:** ▶️ 3 queued — [View all](https://hub.continue.dev/inbox/pr/graphene-data/graphene/103?utm_source=github_pr&utm_medium=pr_body&utm_campaign=continue_tasks)


kevinmarr:
I like that you can use `as` for stored expressions though, because it makes the language feel more consistent and predictable. It's weird if an expression can take a certain form in one place but not another. 

It's cleaner to say that stored expressions are simply SQL expressions that you put in a table statement. Instead of "SQL expressions you put in a table statement except with different aliasing syntax."

Issue #110, opened 2026-02-22
stacked100 charts with date x-axis have scrambled x-axis order


kevinmarr:
## Bug

When using `type=stacked100` on an AreaChart (or any chart) with a date x-axis and a series column, the x-axis renders in a seemingly random order instead of chronologically.

## Repro

In `examples/flights/test-x-axis-sort.md`:

~~~md
```sql weekly_by_carrier
select
  date_trunc('week', dep_time) as week,
  case
    when carrier in ('AA', 'UA') then carrier
    else 'Other'
  end as carrier_group,
  count(*) as flights
from flights
where dep_time >= '2003-01-01'
group by 1, 2
order by 1
```

<AreaChart
  data=weekly_by_carrier
  x=week
  y=flights
  series=carrier_group
  type=stacked100
/>
~~~

The same chart with `type=stacked` renders correctly.

## Root cause

`getStackPercentages` uses `tidy()` from `@tidyjs/tidy`, which strips the `_evidenceColumnTypes` metadata from the data array. In Chart.svelte, `columnSummary` is recalculated after `getStackPercentages` runs (line 305), but now without type metadata, so `getColumnSummary` falls back to `{evidenceType: 'string'}` for the date column.

This causes a cascade:
1. `xDataType` becomes `'string'` → `'category'` instead of `'date'` → `'time'`
2. The sort logic for `category` axes sorts by the **y column value** (descending) rather than the x column — producing the scrambled date order
3. The axis type is set to `'category'` instead of `'time'`, so echarts doesn't auto-sort either

rogueg:
@codex fix this

rogueg:
Yo

PR #118, opened 2026-02-26
Don't drop type info when we use tidy

rogueg:
This was causing axes to render incorrectly when we used tidy to do certain transforms (like stacked100) because the new array didn't preserve the type info.

---

**Continue Tasks:** ▶️ 3 queued — [View all](https://hub.continue.dev/inbox/pr/graphene-data/graphene/118?utm_source=github_pr&utm_medium=pr_body&utm_campaign=continue_tasks)


kevinmarr:
So does this fix https://github.com/graphene-data/graphene/issues/110?

kevinmarr (approved):

PR #121, opened 2026-02-26
Docs refactor

kevinmarr:
- builds entire skill dir in `dist/skills/graphene` instead of just `dist/SKILL.md`
- fixes `name: Graphene` to `name: graphene` (must be lower case to conform with spec)
- breaks graphene.md into sub-files and adds them to subdir `references/` in the skill
- moves best practices out of `cli.md` into its own `best-practices.md`
- updates base.md to be more comprehensive/clear without significantly increasing size (98 lines -> 121 lines)

---

**Continue Tasks:** ▶️ 3 queued — [View all](https://hub.continue.dev/inbox/pr/graphene-data/graphene/121?utm_source=github_pr&utm_medium=pr_body&utm_campaign=continue_tasks)


docs/reference-index.md
rogueg, line 1: Only thing I'd change is to just have the esbuild script construct this dynamically, so you don't have to deal with keeping it up to date.

kevinmarr: oh good idea. i'll update

kevinmarr:
Screenshot 2026-02-26 at 9 53 48 PM
for some reason the rendered markdown shows up like this even though i see them as separate lines in the editor

rogueg:
iirc, markdown requires two newlines to make a new paragraph, or you can use `-` to make it a list. But idk that the rendering matters much, since only agents will read this file, and they just get the text version anyway, not a rendered version.

rogueg (approved):

Issue #108, opened 2026-02-22
Casting numeric/integers to intervals


kevinmarr:
In `example-flights` (on 0.14 as of writing) I've been trying to add this dimension `scheduled_dep_time`, where `dep_time` is the actual departure time and `dep_delay` is the number of minutes delayed:

```sql
scheduled_dep_time: dep_time - interval dep_delay minute
```

This doesn't work in GSQL though. I also tried this, which doesn't work either:

```sql
scheduled_dep_time: dep_time - (dep_delay * interval 1 minute) 
```

I don't know how common this overall use case is, but filing it here for posterity.

dylanscott:
fixed by #123

kevinmarr:
I just tried `scheduled_dep_time: dep_time - interval dep_delay minute` on `flights` on 0.15 and it turns out that that isn't supported by DuckDB. But GSQL happily accepted it and compiled a query that is invalid on DuckDB:

```bash
npx graphene compile "select dep_time, dep_delay, scheduled_dep_time from flights limit 10"

# invalid
SELECT flights."dep_time" as "dep_time", flights."dep_delay" as "dep_delay", (flights."dep_time" - INTERVAL flights."dep_delay" minute) as "scheduled_dep_time" 
FROM flights as flights 
LIMIT 10

# if you run it on DuckDB
Parser Error: syntax error at or near "flights"

LINE 1: ...."dep_delay" as "dep_delay", (flights."dep_time" - INTERVAL flights."dep_delay" minute) as "scheduled_dep_time" FROM...
                                                                       ^
```

Also, the alternative multiplication pattern (from issue above) works on DuckDB but not in GSQL:

```bash
duckdb flights.duckdb "SELECT
  dep_time,
  dep_delay,
  dep_time - dep_delay * INTERVAL '1 minute' AS scheduled_dep_time 
FROM flights
LIMIT 10"
┌─────────────────────┬───────────┬─────────────────────┐
│      dep_time       │ dep_delay │ scheduled_dep_time  │
│      timestamp      │   int64   │      timestamp      │
├─────────────────────┼───────────┼─────────────────────┤
│ 2004-11-18 22:32:00 │        -3 │ 2004-11-18 22:35:00 │
│ 2004-10-12 20:46:00 │         6 │ 2004-10-12 20:40:00 │
│ 2004-11-24 10:20:00 │         0 │ 2004-11-24 10:20:00 │
[...]

PR #123, opened 2026-03-02
support numeric refs in interval expressions

dylanscott:
#108 

Codex (via opencode) 1-shot, looks reasonable to me

---

**Continue Tasks:** ▶️ 3 queued — [View all](https://hub.continue.dev/inbox/pr/graphene-data/graphene/123?utm_source=github_pr&utm_medium=pr_body&utm_campaign=continue_tasks)


lang/analyze.ts
```diff
@@ -601,10 +601,13 @@ export function analyzeExpr (node: SyntaxNode, scope: Scope): Expr {
         if (!parsed) return diag(stringNode, 'Could not parse interval', {sql: 'NULL', type: 'error'})
         return {sql: `INTERVAL ${parsed.quantity} ${parsed.unit}`, type: 'interval'}
       }
-      let num = txt(node.getChild('Number')!)
+      let quantityNode = node.getChild('Number') || node.getChild('Ref')
```
dylanscott, line 604: do you feel strongly about `prefer-const: off` in the lint rules? I find the distinction helpful (mostly to confirm that almost everything is `const`)

rogueg: I disable it out of habit, because I have never had it catch anything useful, and I find having to change vars between let and const to be tedious, but if the agent is going to do it then I suppose I don't care much?

rogueg (approved):

kevinmarr:
hell yeah. nice one

PR #127, opened 2026-03-03
update example AGENTS.md docs links

dylanscott:
I noticed that the examples' AGENTS.md files were pointing to non-existent docs after the refactor in #121. I'm not sure how relevant these are but I asked Codex how it would suggest referencing the refactored docs and this is what it came up with.

---

**Continue Tasks:** ▶️ 3 queued — [View all](https://hub.continue.dev/inbox/pr/graphene-data/graphene/127?utm_source=github_pr&utm_medium=pr_body&utm_campaign=continue_tasks)


kevinmarr:
ah good catch.

kevinmarr (approved):

PR #131, opened 2026-03-03
Fix date_trunc to accept DATE type inputs

kevinmarr:
## Summary
- `date_trunc` on DuckDB and Snowflake rejected DATE columns with `Expected timestamp, got date`, even though both databases natively support DATE inputs
- Added pipe-separated type support (`'date|timestamp'`) to `parseArgType` in `functions.ts`
- Updated arg type for `date_trunc` in `duckDbFunctions.ts` and `snowflakeFunctions.ts`
- BigQuery was already correct (uses generic type `'T'`)

## Test plan
- [x] Added DuckDB and Snowflake test cases to existing `date_trunc` test
- [x] Full `lang` test suite passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

**Continue Tasks:** ▶️ 3 queued — [View all](https://hub.continue.dev/inbox/pr/graphene-data/graphene/131?utm_source=github_pr&utm_medium=pr_body&utm_campaign=continue_tasks)

### DIFF
--- a/lang/duckDbFunctions.ts
+++ b/lang/duckDbFunctions.ts
@@ -1572,8 +1572,8 @@ export const duckDbFunctions: FunctionDef[] = [
     url: `${duck}/date#date_diffpart-startdate-enddate`,
     args: [
       {name: 'part', type: 'string'},
-      {name: 'startdate', type: 'date'},
-      {name: 'enddate', type: 'date'},
+      {name: 'startdate', type: ['date', 'timestamp']},
+      {name: 'enddate', type: ['date', 'timestamp']},
     ],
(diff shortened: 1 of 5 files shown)


rogueg (approved):
could this just be an array instead of the pipe?

kevinmarr:
> could this just be an array instead of the pipe?

@rogueg see latest commit and lmk if that's what you had in mind. if not, maybe you can check out the branch and edit as needed

kevinmarr:
it turns out there were way more functions that had this problem, so i fixed them too

rogueg (approved):

PR #136, opened 2026-03-04
refactor `check` and `run` commands

### LINKED ISSUE #114: Allow agents to see the data behind queries/charts
It would be helpful to the agent to be able to see the raw tabular results of each code-fenced GSQL query in a markdown file. Today, the agent simply uses `run` to iterate over queries, but the issues with this are:
1. The agent tends to introduce drift between the GSQL used in `run` vs. what actually ends up in the markdown, meaning that there could be syntax errors or surprising data that the agent doesn't see directly.
2. The agent may provide a narrative at the end of the turn that is backed up by numbers that are not in the markdown. Which is bad for auditability.
3. The agent has to rewrite the GSQL query with each CLI run, which eats up tokens vs. editing a GSQL query written to a markdown file.

Additionally, I don't think relying on image processing via screenshot is sufficient as a way for the agent to see query results, because:
1. Image processing is far less token/data/inference efficient than processing text
2. The agent tends to not be thorough when processing images - it glosses over a lot of issues

## Solution ideas

I'm not sure what the best approach is for implementation. What I do know is:
- The results should be text, not an image
- We need some way to avoid large query results eating up the context window

Here's one idea, but open to alternatives:
- In addition to saving a .png screenshot in a temp folder, the `check` command also saves a data file (.csv?) for each code-fenced query and tells the agent the path of each file
- Perhaps some metadata is also provided to indicate how large each file is
- Agent will probably use `head`/`tail` etc. as a way to save its own context from large files (not sure if we want to be more strict to prevent the agent from accidentally reading a massive file)
- Questions:
  - When would these files get cleared out?
  - What are the performance concerns?
  - Would need to make sure the user is aware of potentially sensitive data "resting" on their local machine

dylanscott:
This PR refactors the `check` and `run` commands per the discussion in #114. In summary:

- `check` just checks diagnostics, and can be scoped either to a markdown file or a GSQL file
- Taking screenshots is now handled by the `run` command, including the `--chart` flag. Additionally, a `--query` flag can be used to run a code-fenced query from a markdown file.
- **Note**: Contra what was said in #114 I decided _not_ to support `graphene run [ gsqlFile ]`. My reasoning is that it feels counter to the goal of the original issue: The agent should not be iterating on a GSQL query in a separate scratch file, which has the same drift problems. It should ideally iterate on queries in the markdown file/code fence.

Fixes #114.

---

**Continue Tasks:** ▶️ 3 queued — [View all](https://hub.continue.dev/inbox/pr/graphene-data/graphene/136?utm_source=github_pr&utm_medium=pr_body&utm_campaign=continue_tasks)

### DIFF
--- a/cli/check.ts
+++ b/cli/check.ts
@@ -16,7 +16,13 @@ import {pollFor} from '../lang/util.ts'
 import {FILE_MAP} from '../lang/analyze.ts'
 
 interface CheckOptions {
+  fileArg?: string
   mdArg?: string
+  log?: (...args: any[]) => void
+}
+
+interface RunMdFileOptions {
+  mdArg: string
   chart?: string
   log?: (...args: any[]) => void
 }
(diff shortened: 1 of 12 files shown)


rogueg (approved):
Nice! Should we rename some of the functions `run` instead of `check`?

PR #138, opened 2026-03-04
proposal: enable formatting with `oxfmt`

dylanscott:
Hope I'm not being obnoxious but I value having an opinionated code formatter in place. Not because I care a lot about formatting - I don't care much about the specific format at all - but just to completely eliminate it as a thing to think about. This PR enables the `oxfmt` formatter which in addition to being really fast, handles many additional/non code file types out of the box (json, markdown). I'm using the default config with import sorting enabled (using [this default](https://oxc.rs/docs/guide/usage/formatter/sorting.html#example-configuration)). If you're not opposed I will open a PR to give the main/co repo the same treatment.

I was able to get Zed to format on save by adding `"formatter": [{ "language_server": { "name": "oxfmt" } }],` to `settings.json` (and installing the oxc extension). But to try to be unobtrusive I also added a pre-commit hook using lint-staged to format changed files on commit.

---

**Continue Tasks:** ▶️ 3 queued — [View all](https://hub.continue.dev/inbox/pr/graphene-data/graphene/138?utm_source=github_pr&utm_medium=pr_body&utm_campaign=continue_tasks)


rogueg:
Let's talk about it later! 

I def agree that formatting isn't a thing we should need to think about. 

I generally avoid formatters in favor of linters, as I find there are often edge-cases where they make the resulting code much less readable, and `lint --fix` can handle most of the formatting stuff anyway. This is a pretty good article about it: https://antfu.me/posts/why-not-prettier

That said, it does feel like AI is changing things and we should jam on how we want to automate getting code we're happy with.

rogueg:
Ok, played around with it, and I see what you're saying. if you make random whitespace changes, the linter isn't going to do a great job at cleaning those up, beyond fixing indentation.

I'm game to give it a go if it's something you like having! Can we use this config? It's as in-line with our linting rules as I could get it:
```
{
  "$schema": "./node_modules/oxfmt/configuration_schema.json",
  "ignorePatterns": ["docs/**/*.md"],
  "printWidth": 200,
  "semi": false,
  "singleQuote": true,
  "arrowParens": "avoid",
  "bracketSpacing": false,
  "sortImports": {
    "groups": [
      "type-import",
      ["value-builtin", "value-external"],
      "type-internal",
      "value-internal",
      ["type-parent", "type-sibling", "type-index"],
      ["value-parent", "value-sibling", "value-index"],
      "unknown"
    ]
  }
}
```

Would also love if we could disable the pre-commit hook by default. Totally fine to have it checked in, but seems like more of a workflow preference thing. We could add a CI step that ensures that `oxfmt` yields no changes.

Lot of conflicts on this branch, fwiw I found it easier to just roll back to main and re-apply oxfmt 😄

rogueg:
Somewhat related, I've been thinking a lot about how to codify what "good code" looks like at levels above formatting, and it'd be fun to discuss what we think this looks like and how to get agents to do it by default.

dylanscott:
@rogueg sounds good to me. I kept the last commit as the one that actually runs the formatter so it's easy to pop, rebase, and reformat. I did have to disable a couple of lint rules to get them to pass with the reformat:

- `@stylistic/space-before-function-paren`: I think this is an inconsistency for how things are formatted with anonymous vs named functions (it failed when set to `always` or `never` after formatting - I think it wanted a space for anonymous functions but not for named ones?). Seems fine to leave to the formatter
- `@stylistic/indent`: This was failing in only one block of one file (`getColumnExtents.js:15-22`) and the indentation it wanted (in a ternary expression) feels wrong to me. It wanted:

```
    isNumeric
      ? summarize({
        count: n(columnName),
        ...elided,
      })
      : summarize({count: n(columnName), countDistinct: nDistinct(columnName)}),
```

Whereas prettier/oxfmt want:

```
    isNumeric
      ? summarize({
          count: n(columnName),
          ...elided,
        })
      : summarize({count: n(columnName), countDistinct: nDistinct(columnName)}),
```

Also I removed the pre-commit hook, but I already added a CI check so it will block incorrectly formatted code.

rogueg (approved):
sounds great, thanks!

PR #147, opened 2026-03-05
window functions

dylanscott:
This PR adds initial support for window functions to GSQL. It adds a core set of window functions supported across dialects, but does not currently support any dialect-specific window functions (e.g. Snowflake's [ratio_to_report](https://docs.snowflake.com/en/sql-reference/functions/ratio_to_report)).

---

**Continue Tasks:** ▶️ 3 queued — [View all](https://hub.continue.dev/inbox/pr/graphene-data/graphene/147?utm_source=github_pr&utm_medium=pr_body&utm_campaign=continue_tasks)


kevinmarr:
That was fast! My only thing is that there is one line in `base.md` that lists window functions under an "Unsupported" section which you'll want to nix.

Also - general question - do all agg functions accept an optional OVER clause now? What about our special pXX() function?

rogueg (approved):
Looks great!

PR #173, opened 2026-03-10
fanout protection

### LINKED ISSUE #101: Fanout protection
This query is problematic:

```sql
from users select subscription_tier, avg(users.age), sum(orders.price)
```

Because the age gets fanned out by our implicit join of orders, effectively weighting the age by the number of orders placed. It's possible that's the intent, but it's often a footgun.

In gsql, we'd like for such a query to throw an error. If it's intended, we want some kind of language feature that makes it explicit.

dylanscott:
This PR adds a lightweight form of fanout protection as discussed, where GSQL will reject queries (and measure definitions) which use duplicate-sensitive aggregations over different row-multiplying join paths, and forces the agent to rewrite queries to explicitly aggregate at the proper grain. Some choices made in the approach:

1. As mentioned above, we also reject problematic measure definitions in GSQL models as well (e.g. `weird: sum(orders.amount) / sum(payments.amount)`).
2. Adds a `fanoutSafe` override to our function metadata which is used to be smarter about duplication-insensitive aggregates like `MIN`/`MAX`

fixes #101


lang/lang.test.ts
```diff
@@ -1228,6 +1254,111 @@ describe('lang', () => {
     expect('from purchases select customers.name, sum(amount)').toHaveNoErrors()
   })
 
+  it('rejects aggregate queries that mix base and fanout grains', () => {
+    expect('from users select name, avg(age), sum(orders.amount)').toHaveDiagnostic(/Aggregate query mixes fanout localities/i)
```
kevinmarr: Mixing isn't inherently the issue though right? It's that a measure is getting fanned out from an implicit one-to-many join.

For example, this also has a mix but is okay:

```
from orders select users.name, avg(users.age), sum(amount)
```

Could this error message be something like, "Aggregate expression {expr} is fanned out from one-to-many join to {table}. Calculate aggregates in subqueries or CTEs before joining."

rogueg:
Neat! What's the thinking around recovering join metadata? I kinda assumed that if you had a modeled join, you'd always use that and never ad-hoc.

It'd be nice to have a comment somewhere (maybe fanout.ts?) Explaining how this works at a high level. It's not immediately clear to me how fanoutPaths are used.

lang/lang.test.ts
```diff
@@ -914,14 +914,40 @@ describe('lang', () => {
     expect('from users select name, sum(total_orders)').toHaveDiagnostic(/Aggregates cannot be nested/i)
   })
 
-  it.skip('errors if you have a non-agg measure that uses a join_many', () => {
+  it('errors if you have a non-agg measure that uses a join_many', () => {
```
kevinmarr: no such thing as a non-agg measure

lang/lang.test.ts
rogueg: But in that query, won't `avg(users.age)` weight based on the number of orders? Isn't that the kind of thing we want to warn about?

lang/lang.test.ts
```diff
@@ -1228,6 +1254,111 @@ describe('lang', () => {
...
+    expect(`
+      with order_stats as (from users select name, sum(orders.amount) as order_amount),
+      item_stats as (from users select name, sum(orders.order_items.quantity) as item_quantity)
+      from order_stats
+      left join item_stats on item_stats.name = order_stats.name
+      select order_stats.name, order_amount, item_quantity
+    `).toHaveNoErrors()
+  })
+
+  it('rejects aggregate query dimensions that depend on join many', () => {
```
kevinmarr: aggregate query dimension = measure

lang/lang.test.ts
```diff
@@ -1228,6 +1254,111 @@ describe('lang', () => {
...
+    expect('from users select orders.status, sum(orders.amount)').toHaveDiagnostic(/Non-aggregate expressions in aggregate queries cannot depend on a `join many` path/i)
```
kevinmarr: Besides the join to users being useless, I don't see an issue with this query from a fan-out perspective

kevinmarr:
We should discuss this PR and the general join syntax and fan-out stuff at the colo - lots of things to unpack.

My high-level thoughts:
- Often when the agent runs into an error, it simply gives up and pursues something simpler. I worry that when agents run into this issue (particularly on the conservative fall-back case where there is no fan-out) they will give up rather than re-write the query using CTEs.
- AI is getting smarter very quickly. Last time I tried to trick an agent into fanning out aggregates, it didn't fall for it. I think it would be wiser from a product strategy perspective to assume near-infinite intelligence and focus instead on features pertaining to context enrichment or visualization
- We don't have any examples yet of an agent doing bad analysis from a fan out, so maybe we should wait for the problem to exist before fixing
- Point no. 2 above wrt join cardinality recovery strikes me as inelegant (not your fault, it's just a flaw of the current language design)

If we decide not to merge this, I still think the exploration into this was time well spent and was probably also a good ramp exercise :)

Other comments:

>Fields that refer to a `join many` should aggregate

Why? I sense confusion here between aggregation that is materialized into a column vs. on-the-fly aggregation. I know this was a confusion point in the past for us.

>fanout locality

The term "fanout locality" is used in a lot of places here but it's not an industry term and so isn't in any training data. I actually don't know what it means, either... haha. Terms agents will know:
- "Aggregate expression X is fanned out by join to table Y"
- "Join graph creates a chasm trap"

dylanscott:
Re: ad-hoc joins, I was worried since we support them the agent might opt to use them even when the join is modeled, but given that it doesn't sound like that should be an issue I will remove it

dylanscott:
@kevinmarr I took another pass at this to improve the diagnostics and I think they read much more clearly now - let me know what you think.

Also to be clear, the behavior now is that we don't apply any fanout analysis to explicit joins, even if it exactly matches a modeled join (known/modeled join, same join condition). I think that's where we landed?

lang/lang.test.ts
```diff
@@ -1212,6 +1253,112 @@ describe('lang', () => {
...
+      order_stats as (from users select name, sum(orders.amount) as total_amount)
+      from user_stats
+      left join order_stats on order_stats.name = user_stats.name
+      select user_stats.name, avg_age, total_amount
+    `).toHaveNoErrors()
+  })
+
+  it('rejects aggregate queries that mix sibling grains', () => {
+    expect('from users select name, sum(orders.amount), sum(payments.amount)').toHaveDiagnostic(/Join graph creates a chasm trap/i)
+    expect('from users select orders.status, sum(orders.amount)').toHaveDiagnostic(/Expression is fanned out by join to table `orders`; aggregate queries cannot group by it directly/i)
```
kevinmarr: I don't think there is an expression here that is getting fanned out since there are no expressions pointing to user columns

lang/lang.test.ts
```diff
@@ -1212,6 +1253,112 @@ describe('lang', () => {
     expect('from purchases select customers.name, sum(amount)').toHaveNoErrors()
   })
 
+  it('rejects aggregate queries that mix base and joined grains', () => {
+    expect('from users select name, avg(age), sum(orders.amount)').toHaveDiagnostic(/Aggregate expression `avg\(age\)` is fanned out by join to table `orders`/i)
+  })
+
+  it('falls back to a generic diagnostic when a base-grain aggregate mixes with multiple joined grains', () => {
+    expect('from users select name, avg(age), sum(orders.amount), sum(payments.amount)').toHaveDiagnostic(/Aggregate query uses multiple grains \(base, orders, payments\)/i)
```
kevinmarr: Error message should just be chasm trap

lang/lang.test.ts
```diff
@@ -1212,6 +1253,112 @@ describe('lang', () => {
...
+      with order_stats as (from users select name, sum(orders.amount) as order_amount),
+      payment_stats as (from users select name, sum(payments.amount) as payment_amount)
+      from order_stats
+      left join payment_stats on payment_stats.name = order_stats.name
+      select order_stats.name, order_amount, payment_amount
+    `).toHaveNoErrors()
+  })
+
+  it('rejects aggregate queries that mix ancestor and descendant grains', () => {
+    expect('from users select name, sum(orders.amount), sum(orders.order_items.quantity)').toHaveDiagnostic(/Aggregate query uses multiple grains/i)
```
kevinmarr: Could be "Aggregate expression X is fanned out by join to table Y"

lang/lang.test.ts
```diff
@@ -1212,6 +1253,112 @@ describe('lang', () => {
...
+  })
+
+  it('falls back to a generic diagnostic when a base-grain aggregate mixes with ancestor and descendant grains', () => {
+    expect('from users select name, avg(age), sum(orders.amount), sum(orders.order_items.quantity)').toHaveDiagnostic(/Aggregate query uses multiple grains \(base, orders, orders\.order_items\)/i)
```
kevinmarr: "One or more aggregate expressions fanned out by join graph"?

lang/lang.test.ts
```diff
@@ -1212,6 +1253,112 @@ describe('lang', () => {
...
+      with order_stats as (from users select name, sum(orders.amount) as order_amount),
+      item_stats as (from users select name, sum(orders.order_items.quantity) as item_quantity)
+      from order_stats
+      left join item_stats on item_stats.name = order_stats.name
+      select order_stats.name, order_amount, item_quantity
+    `).toHaveNoErrors()
+  })
+
+  it('allows distinct-safe counts to mix with a fanout grain', () => {
+    expect('from users select name, count(id), sum(orders.order_items.quantity)').toHaveNoErrors()
```
kevinmarr, line 1348: Is count() safe because we rewrite to count(distinct)?

dylanscott: Yes, it is. It gets rewritten to `count(distinct id)` here:

https://github.com/graphene-data/graphene/blob/0111890ac99cb2fad4a0ff00d218a3cfd911e4ba/lang/analyze.ts#L464-L471

I added a test case to demonstrate that `count(*)` is treated differently and will hit the fanout protection

lang/lang.test.ts
```diff
@@ -1212,6 +1253,127 @@ describe('lang', () => {
...
+      select order_stats.name, order_amount, payment_amount
+    `).toHaveNoErrors()
+  })
+
+  it('allows aggregate query dimensions when they match the aggregate grain', () => {
+    expect('from users select orders.status, sum(orders.amount)').toHaveNoErrors()
+  })
+
+  it('rejects aggregate query dimensions whose grain does not match the aggregates', () => {
+    expect('from users select orders.status, avg(age)').toHaveDiagnostic(/Expression is fanned out by join to table `orders`; aggregate queries cannot group by it directly/i)
```
kevinmarr: This message doesn't say which expression it's talking about, whereas the one on line 1257 does - possible to make this say "Aggregate expression `avg(age)` is fanned out ..." as well?

kevinmarr (approved):

PR #186, opened 2026-03-12
update GSQL syntax highlighting

dylanscott:
This PR updates the `gsql.tmLanguage.json` grammar uses for syntax highlighting GSQL to fix a couple issues:

1. The biggest one was that the presence of a closed parenthesis in a column type or measure definition would drop out of the `meta.declaration.table.graphene.gsql` scope and break syntax highlighting for the rest of the table. The `#types` and `#functions` patterns were updated to use `begin`/`end` instead of `match` patterns
2. I dropped `measure` as a keyword (which it seemed to be in previous iterations of GSQL, but no longer is), and added `INTERVAL` to the storage type keywords.
3. I also added a test which parses test cases with the grammar and checks the resulting scopes


vscode/gsql.tmLanguage.json
```diff
@@ -116,50 +100,103 @@
       "patterns": [
         {
           "name": "keyword.control.gsql",
-          "match": "(?i)\\b(SELECT|FROM|WHERE|GROUP|BY|HAVING|ORDER|LIMIT|OFFSET|INSERT|INTO|VALUES|UPDATE|SET|DELETE|CREATE|ALTER|DROP|TRUNCATE|REPLACE|VIEW|TABLE|INDEX|PRIMARY|KEY|FOREIGN|REFERENCES|CONSTRAINT|UNIQUE|NOT|NULL|DEFAULT|CHECK|AND|OR|XOR|CASE|WHEN|THEN|ELSE|END|AS|ON|JOIN|INNER|LEFT|RIGHT|FULL|OUTER|CROSS|NATURAL|USING|DISTINCT|ALL|UNION|EXCEPT|INTERSECT|WITH|RECURSIVE|OVER|PARTITION|RANGE|ROWS|BETWEEN|UNBOUNDED|PRECEDING|FOLLOWING|CURRENT|ROW|LIKE|ILIKE|SIMILAR|IN|IS|TRUE|FALSE|UNKNOWN|ESCAPE|ROLLUP|CUBE|FILTER|WITHIN|GROUP|QUALIFY|TOP)\\b"
-        },
-        {
-          "name": "keyword.other.graphene.gsql",
```
dylanscott, line 122: I moved these `JOIN ONE`/`JOIN MANY` cases into the above one, because the `JOIN` keyword was winning due to precedence resulting in the `ONE`/`MANY` not getting highlighted

rogueg:
How hard would it be to add a test that every keyword in the grammar is also in the tmLanguage?

rogueg (approved):

kevinmarr (approved):

Issue #170, opened 2026-03-10
`schema` CLI should emit lower case on Snowflake


kevinmarr:
As of 0.15, on Snowflake the `schema` tool generates table names in upper case ie

```
table ANALYTICS_DATABASE.DATA_PRODUCTS.VIRTUAL_WALKTHROUGH (
  job_id NUMBER
  user_id NUMBER
  session_id TEXT
  org_id NUMBER
...
```

Since the schema tool is used for generating new tables, and because GSQL is case-sensitive on object declarations, this puts you down a path of upper case misery when writing downstream GSQL and dashboards. The current solution is to write a script to lower case all the generated files which is tedious.

Also, the CLI tool itself is case-sensitive which is inconvenient. It would be nice if it translated everything to the database default case unless wrapped in double quotes ie
- `npx graphene schema hEllO` -> shows schema of a database named `hello` in DuckDB and BQ, `HELLO` in Snowflake
- `npx graphene schema "hEllO"` -> shows schema of a database exactly named `hEllO`

Lastly, and least important: Should types be upper case (eg `NUMBER`)? Maybe the schema tool should emit those as lower case too just for sake of consistency and anti-shouting.

PR #194, opened 2026-03-13
Make schema tool case-insensitive

rogueg:
This will find schemas/tables regardless of their actual case in the warehouse, and prints them as lowercase.


kevinmarr:
Fixes https://github.com/graphene-data/graphene/issues/170

kevinmarr (approved):

PR #209, opened 2026-03-19
refactor graphene language server

dylanscott:
This PR refactors the Graphene language server to use the [Volar.js](https://volarjs.dev/) library, and be its own package. At the moment using Volar mostly buys us a slightly nicer framework for organizing LSP functionality, but I thought it was worth trying because it is specifically geared towards supporting embedded languages, like the way we embed GSQL and component syntax in markdown (it is the library both Astro and Vue use to implement their language servers). The `language-server` package supplies a CLI entrypoint so you can run a standalone Graphene language server separate from the VSCode extension.


kevinmarr (approved):

dylanscott:
pretty happy with the code here so I'm going to merge to avoid a stacked PR

PR #210, opened 2026-03-19
update language server to support definitions & references

dylanscott:
This PR builds on the refactored language server from #209 to support go-to-definition and find-references on table and column references in the language server. It also fixes a bug where we were not analyzing column references in join conditions (e.g. `graphene check` would not flag references to columns that don't exist in the model)

### DIFF
--- a/lang/analyze.ts
+++ b/lang/analyze.ts
@@ -18,7 +18,21 @@ import {
 import {analyzeFunction} from './functions.ts'
 import {extractLeadingMetadata} from './metadata.ts'
 import {parseTemporalLiteral, parseIntervalLiteral, parseIntervalUnit, renderTemporalArithmetic, renderStandaloneInterval} from './temporal.ts'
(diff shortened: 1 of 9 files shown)


lang/analyze.ts
```diff
@@ -135,7 +187,12 @@ export function analyzeTableFully(table: Table) {
     c.isAgg = expr.isAgg
     analyzeComputedFieldExpr(c.exprNode, expr)
   })
-  table.joins.forEach(j => (j.table = lookupTable(j.targetNode!)))
+  table.joins.forEach(j => {
+    j.table = lookupTable(j.targetNode!)
+    if (!j.table || !j.onExpr) return
+    let joinTarget = {name: j.alias, table: j.table, alias: j.alias}
+    analyzeExpr(j.onExpr, {table, alias: table.name, fanoutPath: [], joinTarget})
+  })
```
dylanscott, line 195: this is the fix for analyzing column refs in join conditions - without this we wouldn't check columns referenced in join conditions, and wouldn't support go-to-definition on them

kevinmarr (approved):

PR #214, opened 2026-03-20
sync inputs with url

### LINKED ISSUE #189: Bind input values to url params
When you change an input on an app, that value should be put into the url query string, so reloading the page (or sharing) will have the exact same view.

On page load and navigate, we should sync these values from the url into state, rerunning queries if needed.

Relatedly, we should ensure that a table on one page can have a row/cell link attribute that properly populates these params on a different page.

dylanscott:
This PR adds a system to sync dashboard inputs with the URL, so that app state can be shared. There's a fair amount of state management logic in the various input components, but I tried to consolidate the syncing logic into `pageInputs.svelte.ts`

fixes #189

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -32,6 +32,22 @@ export default [
     },
   },
   {
+    files: ['**/*.svelte.ts'],
+    languageOptions: {
+      parser: tseslint.parser,
+      parserOptions: {
+        ecmaVersion: 'latest',
+        sourceType: 'module',
+      },
+      globals: {
+        ...globals.browser,
+      },
+    },
+    rules: {
+      'svelte/prefer-svelte-reactivity': 'off',
+    },
+  },
+  {
     files: ['**/*.svelte'],
     languageOptions: {
       parserOptions: {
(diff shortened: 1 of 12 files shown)


eslint.config.js
```diff
@@ -31,6 +31,22 @@ export default [
...
+        sourceType: 'module',
+      },
+      globals: {
+        ...globals.browser,
+      },
+    },
+    rules: {
+      'svelte/prefer-svelte-reactivity': 'off',
+    },
+  },
```
rogueg, line 49: We already do this just below, I believe

rogueg: Oh, nvm it's svelte.ts files. Bleh what a terrible design choice by the svelte team. Sadly there's no explicit api to tie into their reactivity from plain ts files.

dylanscott: I did end up having to add code to `cli/esbuild.mjs` to handle `.svelte.ts` files, which only manifested in the pack test. Not sure if there's a simpler way to configure that build

kevinmarr (approved):

PR #217, opened 2026-03-23
explicitly pass inputs to analysis pipeline

dylanscott:
This PR refactors `analyze.ts` to remove its use of globals as inputs to analysis (`FILE_MAP` for workspace files, global `config` for dialect, `diagnostics` for outputs). It does not change the analysis logic at all, but just threads this context through. In this PR I have kept a global workspace + result (moved to `core.ts`) as a shim, but would aim to move callers over in a follow-up. Definitely open to suggestions on naming (maybe `AnalysisContext` -> `AnalysisWorkspace`?)

### DIFF
--- a/cli/run.ts
+++ b/cli/run.ts
@@ -7,8 +7,7 @@ import path from 'path'
 import {type PluginOption, type ViteDevServer} from 'vite'
 import {WebSocketServer, type WebSocket} from 'ws'
 
-import {FILE_MAP} from '../lang/analyze.ts'
-import {analyze, config, type GrapheneError, getDiagnostics, loadWorkspace, toSql, updateFile} from '../lang/core.ts'
+import {analyze, config, deleteFile, type GrapheneError, getDiagnostics, loadWorkspace, toSql, updateFile} from '../lang/core.ts'
 import {pollFor} from '../lang/util.ts'
 import {openInBrowser} from './auth.ts'
 import {isServerRunning, runServeInBackground} from './background.ts'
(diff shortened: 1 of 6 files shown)


rogueg:
Slight preference for `Workspace` or `Project` over `AnalysisContext`

This makes analysis stateless, right? The original thinking behind making it stateful is that we could reuse parse and analysis elements between runs to improve performance, which would help in long-running processes like the LSP and server, but could potentially be written to disk as well a la vite.

dylanscott:
Yes this makes the core analysis logic stateless. I think we could still support a more incremental analysis entrypoint, it would just be a more explicit API like `reanalyze(previous: AnalysisResult, updates: AnalysisFileInput[])` or something of that nature (with whatever parsing/analysis artifacts we want to reuse explicitly represented in the result).

IMO it simplifies things to decouple the core analysis logic from that state handling. I could still imagine wanting to share some of that incremental/state logic e.g. between LSP and the watch/serve task, but this way that can be a separate layer rather than being entangled with everything that invokes analysis logic (like how #220 doesn't impact/interact with the CLI path at all)

rogueg:
Let's chat about it today! I feel like a single, stateful flow is going to be easier to reason about than have two different layers, but it's hard to say in the abstract.

Concretely, the previous version reused parse trees between analyses. This would re-parse and re-analyze on every file update, which makes me a little nervous in the LSP case.

dylanscott:
gonna take a fresh pass at this refactor and consolidate PRs

PR #221, opened 2026-03-24
Error cleanups

rogueg:
This code had gotten kinda sloppy and hard to reason about, but it should be simple: when a request fails, it returns an error (serialized as json). That error will always have a message, but in many cases (compilation or analysis, for example) it will have an associated file, maybe a line number, maybe even a code frame for us to render.

### DIFF
--- a/lang/types.ts
+++ b/lang/types.ts
@@ -125,14 +125,21 @@ export interface Position {
   offset: number
   line: number
   col: number
+  lineStart?: number
+  lineText?: string
 }
 
-export interface Diagnostic {
-  file: string
-  from: Position
-  to: Position
+export interface GrapheneError {
   message: string
-  severity: 'error' | 'warn'
+  name?: string
+  stack?: string
+  cause?: unknown
+  severity?: 'error' | 'warn'
+  queryId?: string
+  file?: string
+  from?: Position
+  to?: Position
+  frame?: string
 }
 
 export interface Location {
--- a/package.json
+++ b/package.json
@@ -16,6 +16,8 @@
     "@stylistic/eslint-plugin": "^5.9.0",
     "@types/fs-extra": "^11.0.4",
     "@types/node": "^22.19.7",
+    "@volar/language-server": "^2.4.28",
+    "@volar/language-service": "^2.4.28",
     "dotenv": "^17.2.3",
     "eslint": "^10.0.2",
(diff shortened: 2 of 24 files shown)


tsconfig.json
```diff
@@ -14,5 +14,5 @@
     "types": ["node"]
   },
   "include": ["**/*.ts", "**/*.js", "**/*.svelte"],
-  "exclude": ["node_modules", "cli/dist", "vscode/dist"]
+  "exclude": ["node_modules", "cli/dist", "vscode/dist", "language-server"]
```
dylanscott: I don't think this should be excluded. I refactored the functionality from `languageServer.ts` into this new package with #209, and it consumes the `Diagnostic` type you're refactoring hence the TS errors

rogueg: You're right, I asked an agent to resolve the lint errors from the conflict, and it just ignored the whole dir smh

lang/types.ts
```diff
@@ -162,6 +169,8 @@ export interface FileNavigation {
   references: NavigationReference[]
 }
 
+export type Diagnostic = GrapheneError
```
dylanscott: nit: I think there's only one consumer of this shim export, and it already aliases it when importing it (`language-server/src/service.ts`, aliased to avoid clashing with the VSCode `Diagnostic` type it converts to)

vscode/src/languageServer.ts
```diff
@@ -0,0 +1,155 @@
+import {readFile} from 'node:fs/promises'
```
dylanscott: this is re-adding a file removed with #209. I don't think we want the deps added to the root `package.json` either - `language-server/package.json` declares the necessary ones

package.json
```diff
@@ -25,6 +27,10 @@
     "typescript": "^5.9.3",
     "typescript-eslint": "^8.54.0",
     "vitest": "4.0.18",
+    "vscode-languageserver": "^9.0.1",
+    "vscode-languageserver-protocol": "^3.17.5",
+    "vscode-languageserver-textdocument": "^1.0.12",
+    "vscode-uri": "^3.1.0",
```
dylanscott, line 33: still not sure we need these deps at the root given that `language-server/package.json` pulls in what it uses

dylanscott (approved):

PR #222, opened 2026-03-24
add array support to GSQL

dylanscott:
I thought I'd take a crack at adding array support since it came up with Hover this morning. This PR adds support for `array<T>` syntax in GSQL.

Most of the change here is a refactor of our field type representation from a union of scalars to a tagged union so that we can represent arrays' `elementType`. (I also refactored the new metadata from #228 onto it). I will add `UNNEST` support in a follow-up.


kevinmarr (approved):

kevinmarr:
Approved - re: UNNEST, do you have an opinion on the language design we should pursue?

PR #227, opened 2026-03-30
New viz

rogueg:
Allows configuring echarts directly in md, applies our new styling, and simplifies the Bar/Line/etc chart apis.

The new echarts component lets you do this:
```
<ECharts data="some_table">
  title: {text: "foo"},
  series: [{type: "bar", encode: {x: "month", y: "value"}],
</ECharts>
```

The body of which is standard echarts, so in theory users can make more complex viz.

---
This is **part 1 of 3 in a stack** made with GitButler:
- <kbd>&nbsp;3&nbsp;</kbd> #235 
- <kbd>&nbsp;2&nbsp;</kbd> #234 
- <kbd>&nbsp;1&nbsp;</kbd> #227 👈


kevinmarr:
Does it have to be `<ECharts2>`? Just `<ECharts>` strongly preferred

rogueg:
> Does it have to be `<ECharts2>`? Just `<ECharts>` strongly preferred

Oh, yes it's just `<ECharts>` that was a tmp thing I did while building

rogueg:
@dylanscott @kevinmarr I think this is ready for review.

There's a lot of fast-follow stuff I'd like to do, but I want to avoid this PR getting massive, so I'll start stacking them in other PRs

kevinmarr:
Lgtm! I haven't looked at the implementation yet but I assume it will need some design polish, which I can drop in with my own PR once I'm back.

kevinmarr (approved):

ui/tests/testData.ts
```diff
@@ -102,13 +157,26 @@ export function groupedDataForSection(): TableRows {
...
+  ]
+  return withEvidenceTypes(rows, fields)
+}
+
+function withEvidenceTypes(rows: any[], fields: Field[]): TableRows {
+  ;(rows as any)._evidenceColumnTypes = fields.map(field => ({name: field.name, evidenceType: evidenceType(field)}))
+  return {rows, fields}
+}
+
+function evidenceType(field: Field) {
```
dylanscott, line 174: nit: should we move away from using evidence in names?

ui/component-utilities/types.ts
```diff
@@ -0,0 +1,76 @@
...
+export type SeriesWithGroupingHint = Omit<SeriesOption, 'encode'> &
+  CommonSeriesFields & {
+    stackPercentage?: boolean
+    encode?: SeriesEncode & {
+      group?: string
+      stack?: string
+    }
+  }
+
+export type EChartsConfig2 = Omit<EChartsOption, 'series'> & {
```
dylanscott, line 62: nit: should we remove the trailing `2`?

dylanscott (approved):
LGTM! I definitely think it makes sense for us to own this ourselves, and I like the enrichment approach

ui/tests/testData.ts
rogueg: Yep! This is there temporarily because I haven't touched the non-echarts viz like tables and bigValue. Once we refactor those, we can use the new types entirely.

ui/component-utilities/types.ts
rogueg: Oh good point. Will add it to the list for a future PR

PR #229, opened 2026-03-31
`unnest` support

dylanscott:
This PR adds support for unnesting arrays using an explicit `cross join unnest(field) as alias` syntax. This is closest to BigQuery's syntax, though BigQuery has implicit ["comma cross joins"](https://docs.cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#comma_cross_join), so you can write `from source, unnest(field) as alias`. But keeping the syntax explicit feels right to me for GSQL.


kevinmarr (approved):
Dope!

Issue #156, opened 2026-03-06
Graphene installer


kevinmarr:
At some point it would be good to make the installation process easier and more polished. Our current setup is just a bash script in a Notion doc ([here](https://graphenedata.notion.site/Graphene-setup-instructions-2a75f81b686680a383c4e842a2dc012a?source=copy_link)).

Two ideas, open to more.
- `npm create graphene@latest`, or
- `npx @graphenedata/cli init`

### What it would do:
**1 - Ask the user for all the inputs we need (project name, database type, connection info, etc; see Notion doc above)**
 
Perhaps something like this?
```
┌  Create a new Graphene project
│
◇  Project name
│  my-analytics
│
◇  Database
│  ○ DuckDB (local file)  ● Snowflake  ○ BigQuery
│
◇  Default namespace (optional)
│  MY_DB.ANALYTICS
│
[snowflake branch]
◇  Snowflake account ID
│  myorg-myaccount
◇  Snowflake username
│  graphene_user
◇  Path to .p8 key file
│  /Users/me/.ssh/graphene_snowflake_key.p8
◇  Key passphrase
│  ••••••••
│
[bigquery branch]
◇  GCP project ID
│  my-project-123
◇  Path to service account .json key file
│  /Users/me/.ssh/graphene-bq-key.json
│
[duckdb branch]
◇  Path to .duckdb file
│  ./data.duckdb
│
◇  Describe your data / business (for agent context — edit later in AGENTS.md)
│  We run a SaaS product. Our data lives in Snowflake and covers...
│
◇  Set up coding agent integration?
│  ● Claude Code  ○ Cursor  ○ Windsurf  ○ Skip
│
◆  Installing dependencies...
◆  Setting up agent skill...
└  Done! cd my-analytics && pnpm graphene check
```

**2. Creates the project scaffolding:**
```
my-analytics/
├── package.json          # graphene config embedded here
├── .env                  # KEY_FILE_PATH, PASSPHRASE (non-DuckDB only)
├── AGENTS.md             # or CLAUDE.md
├── .gitignore            # node_modules, *.duckdb, .env
└── tables/               # empty, ready for .gsql files
```

**3. npm install or pnpm install**
Maybe the user does this?

**4. Configures agent skill**
Maybe it uses an npm postinstall script to create the symlink in the path expected for the coding agent?

**5. Adds GSQL files?**
Perhaps we would also add some process for auto-generated .gsql based on the chosen tables or schemas.

PR #236, opened 2026-04-01
graphene npm installer

dylanscott:
This PR creates an installer package for Graphene which (when published) will be able to be invoked with `npm create @graphenedata` (I think the `create-graphene` package is also available on npm - if we grab it we could also support `npm create graphene`). I started with essentially the flow from #156 but we can obviously iterate on that. In the future I think it will be interesting to figure out the next steps (whether or not part of the installer) to get a user started, maybe querying their schema and starting to help build models with them.


create/create.test.ts
```diff
@@ -0,0 +1,159 @@
...
+let introMock = vi.fn()
+let outroMock = vi.fn()
+let cancelMock = vi.fn()
+let isCancelMock = vi.fn((value: unknown) => value === Symbol.for('cancel'))
+let taskLogState = {message: vi.fn(), success: vi.fn(), error: vi.fn()}
+let taskLogMock = vi.fn(() => taskLogState)
+let logErrorMock = vi.fn()
+let spawnMock = vi.fn()
+
+vi.mock('@clack/prompts', () => ({
```
dylanscott, line 22: I don't know how much these heavily mocked tests buy us (it does check some actual generated files). `ui/tests/create-install.test.ts` covers things end-to-end where it actually packages and runs the installer, and validates it can run the created graphene project

cli/connections/duckdb.ts
```diff
@@ -20,13 +20,13 @@ export class DuckDBConnection implements QueryConnection {
   }
 
   private async initialize() {
-    let dbPath = this.options.path
+    let dbPath = this.options.path || config.duckdb?.path
```
dylanscott, line 23: I just added this because #156 showed a flow where you specified a path to a .duckdb file

dylanscott (commented):
I will also need to add handling of the new package to the publish flow

kevinmarr (approved):
Nice! Didn't try it yet but I trust that it works. Would love to grab that package name too.

PR #251, opened 2026-04-14
update metadata syntax

### LINKED ISSUE #124: Change GSQL decorators to `#` instead of `--#`
We already support decorators (via `--#`) as arbitrary metadata but we don't do anything with it, nor do we reserve any particular keys for Graphene use.

**Proposal:**
Do something akin to Malloy's [decorator language](https://docs.malloydata.dev/documentation/language/tags):

```sql
#description: "All orders from e-commerce site" -- optional; comments can still work instead
#synonyms: ["transactions", "purchases"]
#materialization { 
#  type: incremental
#  increment_key: time_col 
#}
table orders (
  ...
  
  #fmt: usd
  #color: green
  total_rev: sum(amount)
)
```

This ought to work inside `select` as well:

```sql
from orders select 
  date,
  #fmt: usd
  #color: green
  sum(amount) as total_rev
```

### Kinds of tags we could support

**Field-level metadata** (base column, dimension, or measure)
- Description (we already support this via inline comments)
- Color (for viz)
- Format (ie. `fmt` value)
- Example values
- Synonyms
- Time truncation level (if DATETIME or TIMESTAMP type; or could be implied for dimensions if expr uses `date_trunc()` or `extract()`)
- Hidden (not `select`-able, but can be used as intermediary logic for other fields)

**Table-level metadata**
- Description (again, we already support this with comments)
- Grain (not sure if this would be just a description of the grain, or an actual tuple of fields)
- Synonyms
- Materialization mode (in the future when we support this)

dylanscott:
This PR updates the ~decorator~ metadata syntax to use a `#key=value` format as discussed. Some decisions made during implementation if we want to change them:

1. Like `--` comments, `#` is treated as line-comments. Only `#key=value` (where value can be quoted) are extracted as metadata. Anything not of that form is ignored. You can include both description and other metadata in a trailing line comment with `-- Description #key=value`
2. Don't currently support a `#flag` shorthand for `#flag=true`. 

Fixes #124


kevinmarr:
For inline decoration I would strongly prefer trailing instead of leading, but won't die on the hill if the parser implementation is too gnarly.

Eg. prefer this
```
my_dim: expr -- This is a comment #tag=val

-- and/or this:

my_dim: expr #tag2=val2 -- This is a comment
```

Instead of this
```
#tag=val my_dim: expr -- This is a comment
```

rogueg:
ooh, yeah I also have a strong gut reaction against the leading comment.

Also would love to continue calling this `metadata` instead of introducing a new term `decorator`.

For parsing, what's the thinking behind moving this into Lezer? Seems like it adds more complexity and restrictions we don't love. The approach we had before just left `--` (and now `#`) as a comment char, and deferred the parsing until metadata.ts.

dylanscott:
@kevinmarr @rogueg that's fair. I've redone the change to be more consistent with `-- description` comments. `#` comments are now line comments from the parser's point of view, and we extract metadata from `#key=value` entries, including trailing inline comments. Like description comments you can apply them with a leading comment, but it has to be on the line above the element

```

#key=value
my_dim: expr 
```

rogueg:
Nice, this is looking good. Should we also add "naked" metadata, like `#hide`?

dylanscott:
@rogueg added naked `#metadata` tags and updated the examples

lang/lang.test.ts
```diff
@@ -737,6 +737,67 @@ describe('lang', () => {
     expect(another.metadata!.units).toBe('seconds')
   })
 
+  it('parses quoted values and multiple hash metadata comments', () => {
+    analyze(`
+      -- currency metrics
+      #color=green #hide #format="US Dollar"
```
kevinmarr, line 743: Two things:
1. What does it mean to format an entire table as `US Dollar`, or to color a whole table as green? I thought that was field-level metadata.
2. Can it use a lower cased ISO 4217 currency code instead of "US Dollar"? Seems more predictable and easier to guess without visiting our documentation

lang/lang.test.ts
```diff
@@ -737,6 +737,67 @@ describe('lang', () => {
...
+      table revenue (
+        amount int,
+        -- gross revenue #units=usd #hide #format="US Dollar"
+        gross int,
+        net int #units=usd #hide #format="US Dollar"
```
kevinmarr, line 748: Oh hm - why does USD have to be specified twice for `units` and for `format`?

lang/lang.test.ts
dylanscott: These are just parser test cases, and I wanted to exercise the case where a value contains whitespace. We have not yet standardized metadata fields and are not using them downstream

kevinmarr (approved):

PR #252, opened 2026-04-14
date part functions

### LINKED ISSUE #174: Support hour(), minute(), dayofweek() and other DuckDB time convenience functions in GSQL
[Claude]

## Context

I was building a dashboard to analyze flight on-time performance by departure hour. My first instinct was to write:

```sql
select hour(dep_time) as dep_hour from flights group by dep_hour
```

This is a valid DuckDB function, but GSQL returned `Unknown function: hour`. I had to fall back to `extract('hour' from dep_time)`, which is more verbose and less discoverable.

## Request

Add support for DuckDB's time convenience functions as aliases in GSQL compilation:
- `hour(ts)` → `extract('hour' from ts)`
- `minute(ts)` → `extract('minute' from ts)`
- `second(ts)` → `extract('second' from ts)`
- `dayofweek(ts)`, `dayofmonth(ts)`, `dayofyear(ts)`, `month(ts)`, `year(ts)`, etc.

These are all valid DuckDB functions and commonly used for time-based grouping. If they can't be supported in GSQL, the error message should suggest the correct `extract()` syntax rather than just saying "Unknown function".

dylanscott:
This PR adds support for shorthand date part-extraction functions as supported natively by duckdb and Snowflake, and lowered to `extract` for BigQuery and the corresponding functions for ClickHouse (`toPart`). We do not try to normalize the behavior of these functions - for parts where behavior varies (mostly week-related) we just pass through to the warehouse. And a `timePart` metadata entry is attached to the field's metadata.

Fixes #174

### DIFF
--- a/lang/analyze.ts
+++ b/lang/analyze.ts
@@ -21,7 +21,7 @@ import {parseMarkdown} from './markdown.ts'
 import {extractLeadingMetadata} from './metadata.ts'
 import {parser} from './parser.js'
 import {parseTemporalLiteral, parseIntervalLiteral, parseIntervalUnit, renderTemporalArithmetic, renderStandaloneInterval} from './temporal.ts'
-import {inferTemporalOrdinal} from './temporalMetadata.ts'
+import {inferTemporalExtractionMetadata} from './temporalMetadata.ts'
 import {
   scalarType,
   type AnalysisConfig,
@@ -1022,11 +1022,10 @@ class AnalysisSession implements Analyzer {
         let unit = txt(node.getChild('ExtractUnit')!)
           .replace(/^['"]|['"]$/g, '')
           .toLowerCase()
-        let timeOrdinal = inferTemporalOrdinal(unit, this.config.dialect)
         return {
           sql: `EXTRACT(${unit} FROM ${expr.sql})`,
           type: scalarType('number'),
-          metadata: timeOrdinal ? {timeOrdinal} : undefined,
+          metadata: inferTemporalExtractionMetadata(unit, this.config.dialect),
           isAgg: expr.isAgg,
           fanout: expr.fanout,
         }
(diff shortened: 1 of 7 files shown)


docs/references/gsql.md
```diff
@@ -298,6 +298,7 @@ Another way of thinking about this is that measures are "self-aggregating."
 - `count` is a reserved word. Do not alias your columns as `count`.
 - Inline window functions are supported using ANSI-style `OVER (...)` clauses. Query-level named windows (`WINDOW w AS (...)`) and set operations (`union [all]`, `intersect`, `except`) are not supported.
 - Percentiles can be computed easily using Graphene's special functions `pXX(col)` (e.g., p50, p975, p9999).
+- Date-part shorthands like `year(ts)`, `month(ts)`, `dayofweek(ts)`, and `hour(ts)` compile across dialects. Calendar semantics such as `week(...)` stay warehouse-native.
```
kevinmarr: What I've been putting in AGENTS.md files is the line "Assume all `<database>` functions are available when writing GSQL queries." So the agent was only attempting these shorthands when connected to databases that supported it.

All that's to say, I'm not sure we need this information in the docs, to save on tokens. Open to arguments to leave this in, too.

rogueg:
What's the word on this? Do we still want these shorthand functions, or do with think leaving it up to the dialect-specific form is fine? From a ux perspective, I could go either way.

If we do want them, I think they _have_ to be in the docs though, right? Otherwise how would the agent ever know to use them?

dylanscott:
DuckDB and Snowflake support them natively, so I think they would fall under the instruction

> Graphene supports almost all functions of the connected data warehouse. Check package.json to see which database you're connected to.

kevinmarr:
Oops, forgot to approve. Minimally we should support them on the dialects that already support them. I think it is unnecessary to support them on the other dialects, however, unless there is no workaround (which in this case, there is).

kevinmarr (approved):

dylanscott:
updated to just catalogue the warehouse-native extraction functions (`year(...)` in DuckDB/Snowflake, `toYear(...)` in ClickHouse)

PR #256, opened 2026-04-15
make warehouse clients peerDependencies of CLI

dylanscott:
This PR updates the CLI such that warehouse clients are peerDependencies so that you only need to install the relevant one, to reduce install size (and the `create` package adds a dependency on the one configured).

### DIFF
--- a/cli/connections/bigQuery.ts
+++ b/cli/connections/bigQuery.ts
@@ -50,6 +50,7 @@ export class BigQueryConnection implements QueryConnection {
   }
 
   async listDatasets(): Promise<string[]> {
+    await this.ready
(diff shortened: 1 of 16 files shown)


kevinmarr (approved):

rogueg:
Can we move the dynamic module loading to connections/index.ts? I feel somewhat strongly that we should keep these connection classes "pure" in the sense that they just deal with connecting to the database based on the given options, not dealing with the context in which they run. That's also why loading env vars into options happens in connections/index.ts.

This matters for `cloud`, where these deps are not optional and we want to load them all up at start, rather than dynamically as we do in the cli.

PR #264, opened 2026-04-15
point CI tests at self-hosted clickhouse

dylanscott:
This PR uses the ClickHouse instance provisioned with #96 for CI testing, to avoid the ClickHouse Cloud premium and flakiness due to instance idling. I re-ran the CI tests here several times and didn't hit any flakes.


dylanscott:
closing in favor of #265

PR #265, opened 2026-04-16
Enable clickhouse tests

rogueg:
And add script for debugging instability I saw while running them, but couldn't root-cause


dylanscott (approved):
Thanks for digging into this! I ran CI several times against the idled cloud instance (with idle timeout set to 5 mins now) and they all passed, so this feels better than going the self-hosted route

PR #267, opened 2026-04-16
rework dynamic connector module loading

dylanscott:
Is this what you had in mind with https://github.com/graphene-data/graphene/pull/256#issuecomment-4256779425?

### DIFF
--- a/cli/cli.test.ts
+++ b/cli/cli.test.ts
@@ -7,6 +7,7 @@ import * as path from 'node:path'
 import {expect} from 'vitest'
 
 import {isServerRunning, stopGrapheneIfRunning} from './background.ts'
+import {preloadWarehouseConnections} from './connections/index.ts'
 
 interface RunResult {
   code: number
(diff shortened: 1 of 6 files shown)


rogueg:
I might be misunderstanding the issue, I kinda figured this would be a noop to the actual connectors like bigQuery.ts. Those files can just directly `import` at the top like they always did, because they're already conditionally imported in index.ts. 

Actually in theory this should require no code changes, but I quite like the better error message you added when the import fails.

dylanscott:
@rogueg I like that. Updated to have `bigQuery.ts`/`duckdb.ts`/etc eagerly import their connectors, and they themselves get lazily imported in `connections/index.ts`

dylanscott:
@rogueg unfortunately actually it looks like that version breaks the pack test. pnpm pack produced a single bundled cli.js where the connector-level static imports were hoisted to the top of the bundle, which made the packaged CLI try to resolve all peer deps at startup

dylanscott:
superseded by #306

PR #270, opened 2026-04-16
Clean up multi-y support

rogueg:
Turns out an agent already did this in an earlier change, but I wanted add tests and clean the code a bit.

### DIFF
--- a/ui/components/ECharts.svelte
+++ b/ui/components/ECharts.svelte
@@ -127,8 +127,9 @@
       let value = queryableEncodeValue(attr, col)
       if (!value) continue
       fields[attr] ||= []
-      fields[attr].push(value)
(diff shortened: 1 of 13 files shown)


ui/components/ECharts.svelte
```diff
@@ -127,8 +127,9 @@
       let value = queryableEncodeValue(attr, col)
       if (!value) continue
       fields[attr] ||= []
-      fields[attr].push(value)
+      if (!fields[attr].includes(column)) fields[attr].push(column)
```
dylanscott: should this be `value`?

dylanscott (approved):

ui/components/ECharts.svelte
rogueg: it should, bad conflict res

PR #274, opened 2026-04-16
Table evidence cleanup

rogueg:
There was a lot of table-only code that I wanted to refactor and clean up while it's top-of-mind.

In particular, there was a lot of logic that isn't really aligned with graphene:
* normalizing to take data in lots of shapes, but with graphene we control the shape.
* trying to infer types by analyzing data. We know the type, and often metadata around that type
* formatting with excel-like format strings. We're dropping this for now, on the assumption that we should be able to format properly on metadata 90% of the time. For the remainder, we'll eventually either add back format strings in some way, or use custom components for this.
* Computed min/sum/max/mean/etc for every numeric column. We pretty rarely need these, it's only for color scales and column totals.


kevinmarr (approved):
Image

you love to see it

dylanscott (approved):

PR #277, opened 2026-04-17
Change chart api to use splitBy and arrange

rogueg:
`<BarChart x=month y=value splitBy=category arrange=group />`

instead of 

`<BarChart x=month y=value group=category />`

Accomplishes more or less the same thing. The main thing that is awkward is if you want to split by multiple dimensions, like a grouped and stacked bar chart. This is no longer possible with the BarChart wrapper, but even with ECharts it's a bit hacky:

Prev that'd be `encode: {x: "month", y: "value", group: "dim1", stack: "dim2"}`
Now it's `encode: {x: "month", y: "value", splitBy: "dim1,dim2"}`

and you just have to to know that in this special case the first one maps to grouping, and the second maps to stacking.

### DIFF
--- a/docs/base.md
+++ b/docs/base.md
@@ -66,11 +66,11 @@ Queries can be referenced by other queries in the `from` or `join` to form DAGs 
 ## Components
 All viz components take `data`, which is the name of a gsql table or code-fenced query in the markdown.
 Component "field" attributes (like x and y) map to a column within data.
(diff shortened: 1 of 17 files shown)


kevinmarr:
Worth mentioning on this ticket that the purpose of this was to make it more intelligible to also support wide data ie y=foo,bar while also having control of grouping or stacking the wide data. Right?

kevinmarr (approved):

PR #281, opened 2026-04-17
Tinker with peer deps

rogueg:
@dylanscott had an agent kick the problem and it came up with this. I think it botched the revert a bit and deleted more than we want, but the last commit looks pretty good. wdyt?

This pr is just a poc, happy to throw it away or use it, whichever you prefer.

### DIFF
--- a/cli/connections/bigQuery.ts
+++ b/cli/connections/bigQuery.ts
@@ -1,11 +1,7 @@
-import type * as BigQueryTypes from '@google-cloud/bigquery'
+import {BigQuery, BigQueryDate, BigQueryTimestamp, type BigQueryOptions} from '@google-cloud/bigquery'
 
 import {config} from '../../lang/config.ts'
 import {type QueryConnection, type QueryResult, type SchemaColumn, type QueryParams} from './types.ts'
-
-type BigQueryModule = typeof BigQueryTypes
-type BigQueryClient = InstanceType<BigQueryModule['BigQuery']>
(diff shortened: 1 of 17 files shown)


dylanscott (commented):
Definitely open to cleaning up the code, but I think if we make the clients peer deps then examples/actual graphene projects explicitly need to pull in their client as a dependency. I think the examples work locally as-is because `cli/` declares the clients as `devDependencies` so Node module resolution is ultimately able to find them, but if you do an isolated install of a Graphene project it will not pull them in, which is why `install.test.ts` is failing. FWIW I had Codex manually run that tests' steps to check the error it's hitting and it says:

> I reran the test outside the sandbox with SLOW_TEST=1 pnpm --dir . exec vitest run ui/tests/install.test.ts --reporter=dot, and it reproduces the same CI shape: the screenshot diff is just the first visible symptom. The real failure
  is the browser console error from /_api/query returning 500.
> I then reproduced the packaged install manually and hit the query endpoint directly. The exact 500 body is
> DuckDB support requires installing @duckdb/node-api. Add it to your project dependencies, for example: npm install @duckdb/node-api
> That lines up with the peer dep change:
> - The scaffold only writes @graphenedata/cli and svelte into the generated app at create/create.ts:176.
> - @duckdb/node-api is now an optional peer dependency in cli/package.json:77.
> So the packaged app installs successfully, graphene check still says “No errors found”, but the first real query execution during page render fails because DuckDB support is missing. If you want, I can patch the scaffold or the install test next.

PR #283, opened 2026-04-17
dry showing tooltips in charts

rogueg:


ui/tests/fixtures.ts
```diff
@@ -161,6 +162,10 @@ export const test = base.extend<{browser: Browser; page: Page; sharedPage: Page;
           el.id = 'component-test'
           container?.appendChild(el)
           window.__inst = window.$GRAPHENE.svelte.mount(window.$GRAPHENE.components[compName], {target: el, props})
+
+          // Wait for load
+          await window.$GRAPHENE?.waitForLoad?.()
+          await new Promise(r => requestAnimationFrame(() => requestAnimationFrame(r)))
```
dylanscott, line 168: out of curiosity, why 2 frames?

dylanscott (approved):

kevinmarr (approved):

ui/tests/fixtures.ts
rogueg: lol at this point it's kinda habit, but the idea is you want to wait until after the ui has actually painted. iirc, the first rAF can happen during the next frame before it paints, but a double rAF will happen after the paint.

PR #285, opened 2026-04-20
non-big <Value/> component

### LINKED ISSUE #32: Implement <Value /> component
The agent does a lot of stuff like this, where it hard-codes numbers it sees from its `run` queries:

Image

Best practice in the real-world would be to keep these values tied to the query like so:

Image

I'll admit this is way less legible (and really a better solution for this use case is to just use a table) but the rendered result looks the same... and this way, the numbers will stay up-to-date as the data changes.

dylanscott:
I noticed this in the backlog and thought I'd give it a shot since I haven't touched the UI components much. This PR adds a simple/unstyled `<Value/>` component which can be used to render query data inline, to avoid hardcoding results in prose. Fixes #32

### DIFF
--- a/docs/base.md
+++ b/docs/base.md
@@ -76,6 +76,7 @@ Attributes like `x`, `y`, `y2`, and `splitBy` are the names of columns within th
 - BigValue: data, value, title, fmt, comparison, comparisonFmt, comparisonTitle, downIsGood, sparkline, sparklineType, sparklineColor
 - Table: data, rows, title, subtitle, groupBy, groupType, subtotals, totalRow, search, sort, link, rowShading, rowNumbers, compact, headerColor
   - Column (sub-component of Table): id, title, fmt, align, wrap, contentType, totalAgg, redNegatives
+- Value: data, column, row
 
 ## ECharts
 To further customize the look and feel of a chart, use the ECharts component to provide an echarts config.
--- a/ui/components/QueryLoad.svelte
+++ b/ui/components/QueryLoad.svelte
@@ -9,13 +9,15 @@
     data: string | QueryResult
     height?: number
     fields?: Record<string, string | string[]>
+    inline?: boolean
     children?: Snippet<[QueryResult]>
   }
 
-  let {data, height = 200, fields = {}, children}: Props = $props()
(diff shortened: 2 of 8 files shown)


dylanscott:
The example from #32 ends up looking like:

````
```gsql top_airplane_models
from flights
select
  aircraft.aircraft_models.manufacturer,
  aircraft.aircraft_models.model,
  count(*) as total_flights
order by total_flights desc
limit 20
```

## Key Findings

**Top 3 Most Active Airplane Models:**

1. **<Value data=top_airplane_models column=manufacturer row=0 /> <Value data=top_airplane_models column=model row=0 />** - <Value data=top_airplane_models column=total_flights row=0 />
2. **<Value data=top_airplane_models column=manufacturer row=1 /> <Value data=top_airplane_models column=model row=1 />** - <Value data=top_airplane_models column=total_flights row=1 />
3. **<Value data=top_airplane_models column=manufacturer row=2 /> <Value data=top_airplane_models column=model row=2 />** - <Value data=top_airplane_models column=total_flights row=2 />
````

=>

Screenshot 2026-04-20 at 9 19 14 AM

kevinmarr:
Love this!

rogueg:
Nice! By convention, most ui tests have at least one screenshot assertion. You could prob get away with just one screenshot and do text assertions for the rest, since this is such a simple rendering.

Do we want and visual hint that this is the result of a query, and not just text? Maybe a subtle underline? cc @kevinmarr 

We might also want to change how we render errors for this one, since QueryLoad by default shows a big block element, but that'd prob be weird for this inline element.

kevinmarr:
In a sense, this is basically a form of grounding, where the convention is to add some kind of visual indicator (eg. in Gemini chat, a little link icon to the web page). However, I could also see situations where you wouldn't want this. What if I have the customer lookup use case and I want to set the H1 header as `<Value>` representing the customer name? Okay then, maybe there would be an attr like `hideLink` or something like that, but now things are getting complicated.

In the past we've also talked about some UI to allow the page viewer to inspect the SQL behind a component. It would make sense to extend this to`<Value>` components as well - so maybe we tackle this then?

kevinmarr (approved):

Issue #258, opened 2026-04-15
ECharts sankey with dataset+encode renders 0 shapes


kevinmarr:
[From Claude]

## Bug

ECharts sankey series do not auto-consume `dataset` + `encode` the way cartesian series do — sankey needs explicit `data` (nodes) and `links` (edges). Graphenes enrich pipeline binds sankey series to `__graphene_base` via `datasetId` + `encode: {source, target, value}`, but at render time `series.getData().count()` is 0 and the chart renders nothing.

This also affects the sankey in `ui/internal/ChartGallery.svelte` (navigate to `/_charts` — sankey panel is blank).

## Repro (examples/flights)

```md
```sql carrier_flow
from flights select
  carriers.name || (pre-9/11) as source,
  carriers.name || (post-2003) as target,
  count() as value
where dep_time >= 2004-01-01
group by 1, 2
```

<ECharts data=carrier_flow height=500px>
  {
    series: [{type: sankey, layout: none, emphasis: {focus: adjacency},
      encode: {source: source, target: target, value: value},
      lineStyle: {curveness: 0.5}}],
  }
</ECharts>
```

Expected: sankey diagram with one node per distinct source/target and one link per row. Actual: empty container.

## Fix

Add an enrichment in `ui/component-utilities/enrich.ts` that materializes `data` (unique node list) and `links` (one per row) from the dataset rows whenever a sankey series only declares `encode.source/target/value`:

```ts
function inlineSankeyData(config, rows) {
  for (let series of config.series) {
    if (series?.type !== sankey || series?.data != null || series?.links != null) continue
    let sourceField = getEncodeField(series?.encode?.source)
    let targetField = getEncodeField(series?.encode?.target)
    let valueField = getEncodeField(series?.encode?.value)
    if (!sourceField || !targetField || !valueField) continue

    let nodeNames = new Set()
    let links = rows.map(row => {
      let s = String(row?.[sourceField] ?? )
      let t = String(row?.[targetField] ?? )
      nodeNames.add(s); nodeNames.add(t)
      return {source: s, target: t, value: Number(row?.[valueField] ?? 0)}
    })
    series.data = Array.from(nodeNames).map(name => ({name}))
    series.links = links
    delete series.datasetId
  }
}
```

PR #292, opened 2026-04-21
Sankey enrichment + example

dylanscott:
This PR adds a `materializeSankeySeries` enrichment, which Codex suggested to massage query results into a format consumable by an ECharts Sankey diagram. I think this is the same issue as #258? With it it was able to create an origin-destination diagram from the NYC taxi data:

Screenshot 2026-04-21 at 9 22 40 AM


kevinmarr:
I'll let @rogueg review this one

rogueg (approved):

rogueg:
Nice!

Issue #299, opened 2026-04-23
`create` script enhancements


kevinmarr:
- Merge the two skills (add subfile and make sure it gets added to the auto-generated index list)
- Detect package manager automatically
- Auto-generate [AGENTS.md](http://agents.md/)
  - “Use [package manager] for all Graphene commands, eg. [package manager] graphene run”
  - “Assume all [database] functions are available when writing GSQL."
- Figure out symlinking (skill installation process)
  - Note: Can’t assume the agent folder (eg. /.claude) already exists
- Update docs/setup.md accordingly
  - Recommend that the human tests the database connection

PR #300, opened 2026-04-23
`create` script enhancements

dylanscott:
This PR does a pass at the `create` command to address most of #299. Specifically:

- It detects the package manager used to invoke the create command, and uses that to install the Graphene CLI (it does not prompt but does still honor an explicit `--no-install` flag).
- It generates a (pretty bare) `AGENTS.md` which gives some example Graphene commands using said package manager. This should be expanded.
- Prompts to symlink the bundled Graphene skill into either `.agents` or `.claude`.

I will handle merging in the modeling skill and any other changes in a separate PR.

### DIFF
--- a/create/cli.ts
+++ b/create/cli.ts
@@ -5,6 +5,7 @@ import {runCreate} from './create.ts'
 await runCreate({
   argv: process.argv.slice(2),
   cwd: process.cwd(),
+  env: process.env,
   stdin: process.stdin,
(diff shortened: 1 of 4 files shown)


kevinmarr (approved):

kevinmarr:
Will you add the line "Assume all [database] functions are available when writing GSQL." to AGENTS.md also?

PR #306, opened 2026-04-24
Refactor where we dynamically import peers

rogueg:
#256 made db sdks peer dependencies, which is great!

But for cloud, they're not peers, and having the async module loading in the connection classes is extra complexity that we don't need.

This is basically the same as #281, but I had an agent do a more targeted revert.


dylanscott (approved):
nice, thanks for cleaning this up

PR #321, opened 2026-04-28
Error on unknown chart props

dylanscott:
I think the most common issue I've had the agent run into is chart misconfiguration due to it passing non-existent props and thinking they are affecting the chart. This PR updates the check command to validate props/attributes passed to our chart wrappers, and includes some prop-specific diagnostics for some common cases (`x/yFmt`, `series` vs `splitBy`). I made these diagnostics errors though I could see a case for making them warnings.

### DIFF
--- a/examples/ecomm/index.md
+++ b/examples/ecomm/index.md
@@ -28,10 +28,10 @@ from order_items select products.category, revenue, gross_profit, gross_margin_p
 from order_items select users.state, revenue, units_sold as units order by 2 desc limit 25
 ```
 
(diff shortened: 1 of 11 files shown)


rogueg:
Yeah, I think telling the agent when it makes this mistake is a good idea.

fwiw, you could also catch this at runtime. iirc, runtime errors are (should be?) returned to `graphene run`, which might be simpler, and would handle mistakes that are harder to statically analyze, as well as custom components.

kevinmarr (approved):

dylanscott:
I think a good case for keeping these as check diagnostics vs runtime errors is that it gets them working in LSP/VS Code for free

Screenshot 2026-04-28 at 1 01 36 PM

PR #323, opened 2026-04-28
Fix table colors to fit with new palette and theme

kevinmarr:
- Map semantic names to Graphene colors
- Fix gradients with `colorScale` (they weren't working)
- Fix a bug where if two <Column>s pointed to the same id, the whole app broke
- Fix the "chip" style delta background color to respect `downIsGood`


rogueg (approved):
lgtm! Added a few tests, and made the bg color computed for color scales.

PR #325, opened 2026-04-29
synthesize `cli/package.json` dependencies

dylanscott:
This PR attempts to improve the dependency management situation with our internal packages - specifically the way `cli/package.json` currently has to re-declare all third-party dependencies used by `ui/` and `lang/`. To that end it adds workspace dependencies from `cli/` to `ui/` and `lang/` for development (but does not update the import patterns - I'm fairly agnostic on whether we should keep relative importing or use internal package imports). For the production publish, a script replaces those workspace dependencies, merging in third-party dependencies to create the published `package.json`. I'm hoping this is a happy medium between the current setup vs. having to publish internal packages to npm.


rogueg:
Not sure I follow, what does this get us?

dylanscott:
> Not sure I follow, what does this get us?

It avoids `cli/package.json` having to re-declare all third-party dependencies pulled in via the internal `lang/` and `ui/` packages

rogueg:
Ah, idk that I mind it terribly. There's a test to ensure it's correct, and it has the advantage of being very simple. But not a hill I want to die on if you feel strongly that we need to synthetically construct it.

dylanscott:
IMO it is worth deduplicating, and avoiding having to know that dependencies need to be added in multiple places

kevinmarr (approved):

PR #329, opened 2026-04-29
LineChart `y2` support

### LINKED ISSUE #304: Support y2 on <LineChart>
`y="series1,series2"` will put them on the same axis, which may not work if they are on wildly different orders of magnitude, or if one is a percent and the other is a total, etc. Need `y2` for two axes.

dylanscott:
This PR adds support for the `y2` prop to the LineChart component, similar to BarChart. This prop was already documented but not actually supported. I also enabled check validation of LineChart props. Fixes #304.


rogueg:
Prob AreaChart too, right?

rogueg (approved):

PR #331, opened 2026-04-29
runtime chart prop checking

### LINKED ISSUE #330: Runtime chart prop checking
#321 added some validation of chart props to to the check command. I think it's worth doing this statically if we can (it e.g. integrates with our LSP/VS Code plugin), but given that it's relatively low-lift we should do runtime validation of props as well (maybe only warning on unknown ones)

dylanscott:
This PR adds runtime checking of extraneous props to our chart components (including `<ECharts/>`), by simply spreading any `...extraProps` we don't use and warning about them. I thought this case should simply warn, since the issue does not block attempting to render the chart. A natural extension of this checking would be to extend the ECharts component to validate the structure of its `config`, though this doesn't look trivial. There does not appear to be an official JSON schema for echarts configuration, but if there are common issues the agent hits when doing raw echarts perhaps we could start building validation based on that.

Fixes #330


cli/run.ts
```diff
@@ -87,21 +87,23 @@ export async function runMdFile(options: RunMdFileOptions): Promise<boolean> {
     return false
   }
 
-  let errors = Array.from(resp.errors || []) as GrapheneError[]
+  let diagnostics = Array.from(resp.diagnostics || resp.errors || []) as GrapheneError[]
```
rogueg: diagnostics are errors, so we shouldn't need this duplication

ui/tests/check.test.ts
```diff
@@ -87,6 +87,119 @@ test('check with mdFile reports unsupported chart wrapper props', async () => {
...
+    <BarChart data="chart_data" x="carrier" y="distance" {...{yFmt: 'num0'}} />
+  `,
+  )
+
+  await page.goto(server.url())
+  let result = await runMdFile({mdArg: 'index.md', log})
+  expect(result).toBe(true)
+  expect(outputLines()).toEqual(
+    trimIndentation(`
+    No errors found 💎
```
rogueg: Seems strange to say "no errors found". There is an error, the agent used a attr that doesn't work.

ui/component-utilities/chartValidation.ts
```diff
@@ -0,0 +1,27 @@
...
+    severity: 'warn',
+    queryId: context || componentName,
+    message: unsupported.message,
+  }))
+}
+
+export function chartContext(componentName: string, data: unknown) {
+  if (typeof data == 'string') return `${componentName} (data="${data}")`
+  return componentName
+}
```
rogueg: Should we just reuse queryId? I do like that this has the component name, we should add that to queryId

rogueg (approved):

rogueg:
Went ahead and fixed this up cause I wanted to cut a release and have this in there, but happy to chat if you think my fix is the wrong direction!

PR #338, opened 2026-05-01
make unknown prop check warn instead of error

dylanscott:
This PR updates the unknown prop checking from #331 to make extra props warnings instead of errors, and prevents them from blocking chart rendering

### DIFF
--- a/cli/check.ts
+++ b/cli/check.ts
@@ -44,11 +44,13 @@ export async function check(options: CheckOptions): Promise<boolean> {
   }
 
   let result = analyzeWorkspace({config, files})
-  if (result.diagnostics.length > 0) {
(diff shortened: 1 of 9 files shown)


rogueg:
I actually think they should be errors. The main user for `run` is agents, and if they've made a mistake, we should make them fix it?

At the moment it's kinda moot. If you add an extra prop, we'll generate a diagnostic and stop `run` before we ever even load the page.
```bash
flights # gp run index.md
ERROR: index.md line 95: Unsupported prop "yFmt" on AreaChart. Use field metadata or ECharts for custom formatting.
    yFmt="foo"
    ^^^^
flights #
 ```

PR #340, opened 2026-05-01
Remove subtitle attribute from BigValue, Table, and PieChart

kevinmarr:
Drop the subtitle prop entirely because agents were overusing it. I also think annotation would be a better feature for this kind of thing. 

Less is more. 

Covers component code, CSS, prop validation,
tests, and docs. The obsolete-prop hint that pointed users to PieChart is
also removed since subtitle is now unsupported everywhere.

https://claude.ai/code/session_01AdqMfYpgkeYore3R6tQe94


rogueg:
> Less is more.

Love it!

rogueg (approved):

PR #342, opened 2026-05-01
Add readme to npm package

rogueg:


kevinmarr:
Oh forgot about npm. I used some GH-specific tags so it will be interesting to see how it renders on npm.

kevinmarr (approved):

PR #345, opened 2026-05-01
Add scatter plot

rogueg:


dylanscott (approved):

kevinmarr:
Nice! One thing about scatters is that it's harder to intuit what the axes are, so they more frequently need the axis titles to make the chart clear. wdyt? The downside is that if we default to showing the axis titles, we will need to make room for those titles which may upset the standard grid positioning we have in our ECharts theme.

rogueg:
Good point, added!

PR #346, opened 2026-05-01
Remove the ability to pass comma-separated lists to ECharts

rogueg:
This sort of snuck in. We want to do this for our wrapper components, but with echarts I think we should stay as close to echarts as possible, and that means manually creating multiple series when you what to show multiple things in wide data.

---

### DIFF
--- a/ui/component-utilities/enrich.ts
+++ b/ui/component-utilities/enrich.ts
@@ -728,14 +728,13 @@ function getSeriesValueFields(series: SeriesWithGroupingHint, fields: Field[]) {
   }
 }
 
-// The props on series.encode can either be a string, an array, or a comma-delimited string.
+// The props on series.encode can either be a string, or an array.
 // In all cases, this returns the corresponding fields for each item.
 function getEncodeFields(series: SeriesWithGroupingHint | undefined, fields: Field[], encodeProp: string): Field[] {
   let raw = series?.encode?.[encodeProp]
-
   let names: string[] = []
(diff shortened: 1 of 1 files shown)


rogueg:
Please push back if you think we should have this. I could be convinced.

dylanscott (approved):
This makes sense to me - we should be able to link to the ECharts docs as context for the `<ECharts/>` component without adding too much on top

kevinmarr:
Seems inconsistent with supporting `splitBy` though, which is also not an ECharts thing and could be accomplished with multiple series.

In our thin wrappers, `splitBy` and arrays in `y` handle [long-form data and wide-form data](https://www.statology.org/long-vs-wide-data/) respectively, so feel like complementary features.

So seems logically inconsistent to support one but not the other for ECharts, if neither are native ECharts features.

rogueg:
Yes, exactly. Good point. It's logically inconsistent. 

The thing I will say though is that echarts has no way to automatically create series for tall data, hence the addition of `splitBy`. It does have an easy and common way to handle wide data by creating multiple series. If we added comma x/y, we'd be telling the agent "hey, don't do this the way you're used to doing it with echarts, do it this other non-standard way instead"

So it's logically inconsistent either way 😄

rogueg:
cc @kevinmarr

kevinmarr (approved):

kevinmarr:
i buy it - approved

PR #356, opened 2026-05-04
Prevent commits from unexpected contributors

rogueg:


dylanscott (commented):
It feels weird to hard-code us as the the only three authors in a check like this - would it be reasonable to instead blacklist our known-wrong emails (dylan.scott@gmail.com for me)?

Perhaps more importantly, this job would not have blocked any of the commits in the repo I made with my personal email. All of the commits I've created locally have been with my graphene email, but I didn't realize that in the GitHub UI when I click "squash and merge" it has been defaulting to my personal email, which is too late for this check:

Image

I'm hoping that changing my primary email to dylan@graphenedata.com has fixed this but I haven't been able to confirm yet (it didn't change it on an already-open PR) EDIT: I think the default change did take now 🙏

rogueg:
It wasn't just our personal emails, there were things from `pi`, `continue` in the author field for some commits. An allowlist catches all the stuff we're not expecting, in addition to personal emails.

Re github setting the author when you squash-merge, idk if there's a way to enforce this in the github ui. There is an enterprise feature around commit metadata, so I suppose we could try that if we wanted to.

dylanscott (approved):

PR #358, opened 2026-05-04
input-aware execution for `graphene run`

dylanscott:
This PR updates `graphene run` to take into account some feedback from using graphene to build an input-heavy dashboard. There were a couple of turns where the dashboard was just completely broken on load, with all components failing. This turned out to be due differences between the result of `graphene run` vs an actually-loaded dashboard with input values. This PR updates the run command to make several changes:

1. By default queries will be run with the default values from their corresponding input components, but can be overriden with `--input`
2. The run command supports a `--list-inputs` flag to print static page inputs, defaults, and locations
3. The run command has a new `--all-queries` flag which will run every query on the page headlessly

This did require refactoring some default-handling code from svelte components into more reusable helpers, but nothing too crazy.

### DIFF
--- a/cli/cli.test.ts
+++ b/cli/cli.test.ts
@@ -54,6 +54,20 @@ async function createTelemetryProject(prefix: string) {
   return tmpDir
 }
 
+async function createFlightProject(prefix: string) {
+  let tmpDir = await fsp.mkdtemp(path.join(os.tmpdir(), prefix))
+  let pkg = {
+    name: prefix,
+    graphene: {
+      dialect: 'duckdb',
+      duckdb: {path: path.join(flightDir, 'flights.duckdb')},
+    },
+  }
+  await fsp.writeFile(path.join(tmpDir, 'package.json'), JSON.stringify(pkg, null, 2) + '\n')
+  await fsp.cp(path.join(flightDir, 'tables'), path.join(tmpDir, 'tables'), {recursive: true})
+  return tmpDir
+}
+
 describe('cli compile', () => {
   it('compiles a basic query (happy path)', async () => {
     let res = await runCli(['compile', 'from flights select carrier'], {cwd: flightDir})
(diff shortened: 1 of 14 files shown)


kevinmarr:
Ah good catch! 

I'm confused about this:

>The run command has a new `--all-queries` flag which will run every query on the page, including unnamed queries derived from viz components

What is the behavior without the flag? I assumed all queries were always run

dylanscott:
Yeah actually pulling on that thread with the agent is interesting: the agent hit a lot of errors with the browser automation layer that a full `graphene run page.md` command invokes, and as a result ended up falling back to validating only with `run` commands that would work headlessly. From Codex:

>   **Browser/socket validation is brittle for agents**
  The full page runner uses a browser tab registration flow. The server exposes `/_api/check`, but when I posted to it, it returned:
>
>  `{"error":"no_tab"}`
>
>  even though lsof showed a browser process connected to port 4000.
>
>  That likely means the browser connection existed at the TCP level, but Graphene did not have a registered websocket for the exact normalized page URL. This kind of stateful browser bridge is fine for “take a screenshot of what I’m looking at,” but it’s fragile as an automated correctness check.
>
>  For agents and CI, the validation path should not require:
>
>  - opening a browser,
>  - finding the right tab,
>  - a websocket registration,
>  - GUI permissions,
>  - local port probing from a sandbox.

So this is more about exposing more runtime checking headlessly. I think that makes sense but maybe there's more we can do to get the browser-based checking to work as-is

kevinmarr:
Gotcha - I'm hazy on the details but I do remember Grant having to wrestle this a bit to get the screenshots working, so I would defer to him for the PR review.

rogueg:
What were the differences between between `graphene run` and a loaded dashboard?

I'm curious what errors the agent ran in to using `run`. Those sound like bugs we should fix.

I like the idea of adding `--input`, and because of the work you did binding input state to the url, this should actually be a pretty easy change, as you can just set the url!

dylanscott:
Digging deeper through the agent logs it looks like the main thing was that it was intermittently hitting issues with `graphene run` like

```
Starting Graphene server...
file:///Users/dylan/Code/example-nba/node_modules/@graphenedata/cli/dist/cli/chunk-OVWODUTJ.js:12200
      reject(new Error("Exited before server started"));
             ^

Error: Exited before server started
    at ChildProcess.<anonymous> (file:///Users/dylan/Code/example-nba/node_modules/@graphenedata/cli/dist/cli/chunk-OVWODUTJ.js:12200:14)
    at Object.onceWrapper (node:events:623:26)
    at ChildProcess.emit (node:events:508:28)
    at ChildProcess._handle.onexit (node:internal/child_process:294:12)
```

And eventually gave up on trying to use it for validation. I'm going to take a fresh pass at the input side of this change because I think that is valuable, but I was somewhat confused going into this and I think it can be trimmed down. And I will try to do some more runs to see if I can repro the server issue it was hitting

rogueg:
Interesting. When you `run`, it will attempt to start up the server in the background if one isn't running. This error you're seeing I think happens when it fails to start up that server.

It seems like we're swallowing whatever error actually stopped the server from starting up, but you might be able to find it by looking at `node_modules/.graphene/serve.log`

dylanscott:
Good call - maybe a sandbox issue?

```
10:32:46 AM [vite] (client) Re-optimizing dependencies because lockfile has changed
node:net:1926
      const error = new UVExceptionWithHostPort(rval, 'listen', address, port);
                    ^

Error: listen EPERM: operation not permitted 127.0.0.1:4000
    at Server.setupListenHandle [as _listen2] (node:net:1926:21)
    at listenInCluster (node:net:2005:12)
    at node:net:2214:7
    at process.processTicksAndRejections (node:internal/process/task_queues:90:21) {
  code: 'EPERM',
  errno: -1,
  syscall: 'listen',
  address: '127.0.0.1',
  port: 4000
}

Node.js v24.14.0
```

PR #359, opened 2026-05-05
add `--input` option to `graphene run`

dylanscott:
Taking another pass at this, much simplified from #358. This PR adds a `--input` option to `graphene run` which can be used to specify input values. This is particularly important for running parameterized queries with `graphene run -q` which previously you couldn't do. This pass I opted to avoid all the bits related to listing inputs and extracting default input values. That was kind of nice, but it added a lot of complexity and I think this should work as an easy loop if the agent hits an "Error: missing param $foo" when attempting to run a query.

### DIFF
--- a/cli/cli.test.ts
+++ b/cli/cli.test.ts
@@ -116,6 +116,13 @@ describe('cli run', () => {
     expect(res.stdout.toLowerCase()).toContain('total')
   })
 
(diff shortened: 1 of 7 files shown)


kevinmarr:
Don't forget to update docs :)

kevinmarr (approved):

rogueg (approved):
nice!

PR #362, opened 2026-05-05
rework `<GrapheneQuery/>` prop escaping to avoid Svelte interpolation

dylanscott:
This turned out to be an issue with the rendering of the GrapheneQuery svelte component - if you ran `npx graphene run repro-gt-escape.md -q delayed_flights` it would show you the actual duckdb error with the format string. However the `escapeHtml` helper used to escape query when compiling markdown was not escaping everything that the Svelte parser would interpret. Per Codex:

>  cli/mdCompile.ts was turning the fenced SQL into:
>
>  ```
>  <GrapheneQuery code="select
>    format('{:.1%}', 0.5) as pct
>  from flights" />
> ```
>
>  Svelte still parses {...} inside quoted component attributes as JS interpolation, so {:.1%} was treated as invalid JS before the query ever reached DuckDB.

~I opted to fix this by simply base64 encoding the name/code props rather than trying to escape them. This feels a bit gross, but trying to fully escape these in a way that definitely survives the full mdsvex/rehype/sanitize-html pipeline looked hairy.~

Per your comment, reworked to use the approach from https://github.com/sveltejs/svelte/issues/2602#issuecomment-487375068 to avoid the Svelte interpolation

With the change the duckdb error bubbles up as expected:
Screenshot 2026-05-05 at 2 07 32 PM

### DIFF
--- a/cli/mdCompile.test.ts
+++ b/cli/mdCompile.test.ts
@@ -35,6 +35,25 @@ from flights
     expect(code).toContain('<GrapheneQuery name="{`repro`}" code="{`select\\n  format(\'{:.1%}\', 0.5) as pct\\nfrom flights`}"></GrapheneQuery>')
   })
 
+  it('keeps query comparison operators as JavaScript escapes instead of HTML entities', async () => {
+    let src = `
+\`\`\`sql repro
+select * from products
+where created_at >= coalesce($daterange_start, created_at)
+  and created_at <= coalesce($daterange_end, created_at)
+\`\`\`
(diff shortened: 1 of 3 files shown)


rogueg:
I didn't even know svelte did interpolation in strings. What a weird feature, which you already have js string interpolation. An alternate fix: https://github.com/sveltejs/svelte/issues/2602#issuecomment-487375068

rogueg (approved):

PR #368, opened 2026-05-06
fix LSP `watchFiles` registration

dylanscott:
This turned out to just be an issue with where we were registering our file watcher. By the time `watchWorkspace` is called (in the service's `create`), the initialization callbacks have already fired, and so the file watcher logic in the `server.onInitialized` callback was never invoked. This PR updates the setup path to simply register the watcher directly on that path rather than using `server.onInitialized`, since the server is already initialized.

Tested with the `aircrafts.gsql` repro - both clearing the file and deleting it altogether - and both cases update the diagnostics for other files without `aircrafts.gsql` being open.

### DIFF
--- a/language-server/src/service.test.ts
+++ b/language-server/src/service.test.ts
@@ -2,10 +2,10 @@
 import {mkdtemp, mkdir, rm, writeFile} from 'node:fs/promises'
 import os from 'node:os'
 import path from 'node:path'
-import {expect} from 'vitest'
+import {expect, vi} from 'vitest'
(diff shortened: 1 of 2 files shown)


rogueg:
Oh nice this looks much cleaner! I'm still ambivalent about volar, but battle for another day I think.

Would be nice to have extension tests that were a bit more e2e. What do you think about https://code.visualstudio.com/api/working-with-extensions/testing-extension

kevinmarr (approved):

PR #377, opened 2026-05-07
[release] 0.0.17

rogueg:

PR #379, opened 2026-05-07
fix interaction between release PR and install.test.ts

dylanscott:
test-install is failing on #377 due to a funny interaction where it tries to resolve the not-yet-released version. Interestingly only pnpm fails in this way - the npm and yarn tests pass and seem to realize they are substituting the tarball installed before trying to resolve the version. This PR updates the install test to explicitly swap in the `@graphenedata/cli` dep in package.json with a `file:` dependency pointing to the tarball to avoid this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,24 @@
+## 0.0.17
+
+### Breaking changes
+
+None.
+
+### Added
+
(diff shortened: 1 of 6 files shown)


dylanscott:
closing in favor of the fix added in #377, which is cleaner

PR #383, opened 2026-05-08
make `graphene run page.md` run server in foreground

dylanscott:
This PR updates the `graphene run page.md` flow to spawn the graphene server in the foreground (in the same way as `graphene serve` without `--bg`). Unfortunately this doesn't actually affect the interaction with the default Codex sandbox behavior (it does not allow localhost binding by default, whereas Claude Code does - see #384). But it does change the error from the run command in that case from the very cryptic

```
Starting Graphene server...
file:///Users/dylan/Code/example-nba/node_modules/@graphenedata/cli/dist/cli/chunk-OVWODUTJ.js:12200
      reject(new Error("Exited before server started"));
             ^

Error: Exited before server started
    at ChildProcess.<anonymous> (file:///Users/dylan/Code/example-nba/node_modules/@graphenedata/cli/dist/cli/chunk-OVWODUTJ.js:12200:14)
    at Object.onceWrapper (node:events:623:26)
    at ChildProcess.emit (node:events:508:28)
    at ChildProcess._handle.onexit (node:internal/child_process:294:12)
```

to:

```
Starting Graphene server...
Failed to start Graphene server
listen EPERM: operation not permitted 127.0.0.1:4000
```

### DIFF
--- a/cli/cli.test.ts
+++ b/cli/cli.test.ts
@@ -155,6 +155,16 @@ describe('cli run', () => {
     expect(res.stdout).toContain('AA')
   })
 
+  it('stops an owned markdown server after a no-browser run', async () => {
+    await stopGrapheneIfRunning()
+
+    let res = await runCli(['run', 'index.md'], {cwd: flightDir})
+    expect(res.code).toBe(1)
+    expect(res.stdout).toContain('Starting Graphene server')
+    expect(res.stdout).toContain('Failed to open a new tab')
+    expect(await isServerRunning(TEST_PORT)).toBe(false)
+  })
+
   it('prints a clean error for a missing markdown query input', async () => {
     let res = await runCli(['run', 'carrier_detail.md', '--query', 'carrier_info'], {cwd: flightDir})
     expect(res.code).toBe(1)
--- a/cli/run.ts
+++ b/cli/run.ts
@@ -13,7 +13,7 @@ import {analyzeWorkspace, getFile, loadWorkspace, toSql} from '../lang/core.ts'
 import {type AnalysisResult, type WorkspaceFileInput} from '../lang/types.ts'
 import {pollFor} from '../lang/util.ts'
 import {openInBrowser} from './auth.ts'
-import {getGrapheneCache, isServerRunning, runServeInBackground} from './background.ts'
+import {getGrapheneCache} from './background.ts'
 import {runQuery} from './connections/index.ts'
 import {mockFileMap} from './mockFiles.ts'
 import {normalizeFile} from './normalizeFile.ts'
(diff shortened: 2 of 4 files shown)


kevinmarr (approved):

rogueg:
This feels off to me. If you're doing local work with Graphene, you'll likely want to be able to look at the results yourself. It's weird for the agent to say "hey I built this dashboard to answer your question, but you can't see it".

For that, you need the server running. If an agent can't launch that server for you (either because of the sandbox, or because agents have historically been bad about working with long-running processes), maybe we should just ask the user to run it themselves?

What's in the stdout/stderr from the spawn? I'd guess it's the same error message, and we can just show that to the user. Or better still, we could detect this _specific_ type of error and provide instructions on exactly how to recover.

dylanscott:
To be honest it felt pretty unexpected to me that `graphene run page.md` would spawn a persistent background `graphene serve` process that outlives the `graphene run` process. Do we document that behavior? It can cause some weird interactions, especially with the `isServerRunning` check in `run.ts`, where if I have one of those serve processes from a different graphene project hanging around, a `graphene run` on a different project will think it is already running and screenshot a tab of the other project.

kevinmarr:
Good topic for discussion today. I find it disruptive that the agent's work opens tabs in my browser and disrupts the flow of whatever I'm doing, but I'm not sure if that behavior is independent from what you're describing here.

My ideal would be:
1. I ask the agent to do some work on a page
2. If using the Codex/CC/Cursor desktop app, all those tools have a built-in browser and the agent opens that up automatically to show me the work its done. It works irrespective of whether I have the dev server running already. It does NOT open tabs in my own browser.
3. If not using a desktop app with that capability, when the agent is finished I can spin up the dev server myself and view the page in my browser. While it would be nice as the user to not worry about spinning up the dev server, if the tradeoff is that the agent always leaves behind a persistent dev server running in the background I'm not sure that's worth it.

rogueg:
The original thinking was to make this easy for less-technical data folks. You could ask the agent "hey build me a report on blah", it'd think for a bit, and come back saying "All done! You can check it out at http://localhost:5000/blah". You don't have to know anything about running a server in a different terminal tab. I agree that it's a little unusual if you're technical, and used to starting up a dev server to do work.

Malloy (and Evidence) built the server running into the vscode extension, but I don't think we want to be dependent on an ide.

The question of which browser to use is kinda separate. Seems like our choices are:
* no browser, for agents that have their own
* user's browser, low overhead, but it opens a tab
* playwright, have to download/run a completely separate browser

I think you also bring up a good point of how graphene behaves if you have multiple projects running at the same time. We could do more to ensure the cli is talking to the right server, and to not kill the wrong server. I'd also guess it's going to be pretty rare in the early days.

rogueg:
Oh, there's a 4th option that's kinda related to both: no server or browser. It might be possible purely in js to render the page and rasterize charts. You'd might end up with a lot of discrepancies for html stuff, and wouldn't be able to support the full range of what custom components might do, but for pure echarts it might be possible.

dylanscott:
closing this in favor of documenting background behavior, and adding logic to tail the serve log to give better errors

PR #384, opened 2026-05-08
have create-graphene write Codex project config

dylanscott:
It turns out the issue with `graphene run page.md` in Codex is very simple: Codex's default sandbox policy disallows even localhost binding, whereas Claude's allows it. This makes the difference between a page run working and not.

This PR updates the create-graphene flow to adjust how we prompt for agent setup (asking which agent to set up, with the options being "Claude", "Codex", or "Other agent"). In the Codex case we write a `.codex/config.toml` in the project folder with a default policy that enables `allow_local_binding`. In my opinion this feels totally above board - we are writing a Codex config scoped to the generated project, and only adjusting the localhost network access behavior (to be like Claude's default).

### DIFF
--- a/create/create.test.ts
+++ b/create/create.test.ts
@@ -304,6 +304,7 @@ describe('runCreate', () => {
     expect(spawnMock).toHaveBeenCalled()
     expect(await lstat(path.join(root, 'demo-app/.agents/skills/graphene')).catch(() => null)).toBeNull()
     expect(await lstat(path.join(root, 'demo-app/.claude/skills/graphene')).catch(() => null)).toBeNull()
+    expect(await lstat(path.join(root, 'demo-app/.codex/config.toml')).catch(() => null)).toBeNull()
   })
 
   it('does not prompt before installing dependencies', async () => {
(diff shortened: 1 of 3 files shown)


kevinmarr:
Sounds good to me. Do other standard dev flows with Vite not work in Codex either? Or is it something more specific to Graphene?

kevinmarr:
Would be worth testing Cursor and OpenCode too unless it's too much trouble

dylanscott:
> Do other standard dev flows with Vite not work in Codex either? Or is it something more specific to Graphene?

I was very surprised this is the default in Codex because it is nothing specific to Graphene - you would have to customize your sandbox settings for any webdev flow where you want the agent to be able to run a server. But it's easy to confirm:

> › I want to test something about your default sandbox settings. Can you try to bind an HTTP server to any port on 127.0.0.1, maybe with a Python one-liner?
>
> • I’ll run a small local bind test with Python from the current workspace. It will only listen on loopback briefly and report whether the sandbox allows it.
>
> • The default sandbox blocks binding a local HTTP server to 127.0.0.1.
>
>   Command attempted: Python `socketserver.TCPServer(("127.0.0.1", 0), ...)`
>   Result:
>   `PermissionError: [Errno 1] Operation not permitted`

As far as I can tell OpenCode does not come with an included sandbox implementation, and expects you to take care of it by running in a container or whatever. I can try Cursor

kevinmarr (approved):

PR #385, opened 2026-05-08
Give mtoy some love!

kevinmarr:
He also deserves credit and he is a gem of a guy :) Fun fact: also the co-creator of [Rogue](https://en.wikipedia.org/wiki/Rogue_(video_game)).


dylanscott (approved):
wow, legend! Roguelikes have become a genre unto themselves

Issue #23, opened 2025-09-11
Ability to hide columns via tags


kevinmarr:
Hiding makes it so that the consumer GSQL SELECT query on that model cannot see or query the hidden object, with the main benefit being to clean things up, declutter, and reduce confusion.

Why not just delete or comment things out?
* You can't delete database columns in Graphene (nor should you)
* Intermediary logic: Sometimes you want to declare modeled objects solely to be accessible to other modeled objects, as a means of DRYing up logic or making it more legible, but don't want or need it exposed to consumer GSQL queries

Should support
* Same line annotation (trailing)
* Preceding line(s) annotation
* Combination with comments
* Ability to hide arbitrary expressions and joins

```
extend my_table (

  column_to_hide, # hide

  column_to_hide2, -- A description # hide

  # hide
  -- A description
  column_to_hide3,

  -- A description
  # hide
  column_to_hide4,

  # hide
  1 + 1 as two,

  # hide
  sum(two) as even_num,

  # hide
  join one another_table on ...
);
```

I thought about the ability to hide entire tables, but it got really complicated and I'm not sure it even makes sense if GSQL will be able to query un-modeled database tables.

PR #386, opened 2026-05-08
validate metadata in `check`

dylanscott:
This feels like a good opportunity to do a pass on our supported metadata. Personally I find it _very_ weird for us to use `#units` to specify currency (as well as physical units? At least there was an existing example in `lang.test.ts` with `#units=seconds`. It's also strikes me as weird that it's pluralized, it should be `#unit`). Could we use a dedicated `#currency` metadata, and reserve `#unit` for physical dimensions?

### DIFF
--- a/docs/references/gsql.md
+++ b/docs/references/gsql.md
@@ -147,7 +147,7 @@ For this reason, some metadata should be set explicitly in the GSQL model, using
 | `#ratio` | no | Value is 0–1; rendered as `value × 100%` (e.g. `0.42` → `42%`) |
 | `#pct` | no | Value is already 0–100; rendered as `value%` (e.g. `42` → `42%`) |
 | `#currency=<code>` | no | Adds currency symbol and compacts to K/M/B. Accepts ISO 4217 currency codes like `USD`, `EUR`, and `JPY` |
-| `#unit=<unit>` | no | Physical unit label. Any non-empty value is accepted |
+| `#unit=<unit>` | no | Appends the provided value to the end of labels in visualizations (e.g. `unit=minutes` appends "minutes", or "(minutes)" on axes). Any non-empty value is accepted |
 | `#timeGrain=<grain>` | yes (from `date_trunc`, `date_bin`, casts) | Controls time axis label format. Values: `year`, `quarter`, `month`, `week`, `day`, `hour`, `minute`, `second` |
 | `#timeOrdinal=<ordinal>` | yes (from `extract`) | Treats values as ordinal positions on a categorical axis. Values: `hour_of_day`, `day_of_month`, `day_of_year`, `week_of_year`, `month_of_year`, `quarter_of_year`, `dow_0s` (0=Sun), `dow_1s` (1=Sun), `dow_1m` (1=Mon) |
 | `#description=<text>` | no | Description text for a table or field. `--` comments are also collected as descriptions |
--- a/docs/references/gsql.md
+++ b/docs/references/gsql.md
@@ -151,10 +151,7 @@ For this reason, some metadata should be set explicitly in the GSQL model, using
 | `#timeGrain=<grain>` | yes (from `date_trunc`, `date_bin`, casts) | Controls time axis label format. Values: `year`, `quarter`, `month`, `week`, `day`, `hour`, `minute`, `second` |
 | `#timeOrdinal=<ordinal>` | yes (from `extract`) | Treats values as ordinal positions on a categorical axis. Values: `hour_of_day`, `day_of_month`, `day_of_year`, `week_of_year`, `month_of_year`, `quarter_of_year`, `dow_0s` (0=Sun), `dow_1s` (1=Sun), `dow_1m` (1=Mon) |
 | `#description=<text>` | no | Description text for a table or field. `--` comments are also collected as descriptions |
-| `#hide` | no | Marks a table or field as hidden. |
 | `#pii` | no | Marks a field as containing personally identifiable information. |
-
-Flag annotations (`#ratio`, `#pct`, `#hide`, `#pii`) accept either bare usage or `=true`.
 
 ## `select` statements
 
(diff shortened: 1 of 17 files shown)


docs/references/gsql.md
```diff
@@ -138,18 +138,20 @@ There isn't always a SQL expression that can tip Graphene to the semantic meanin
 - The field could be a base column that has no source expression
 - There might not be enough information in the expression (eg. what currency a float is tied to)
 
-For this reason, some metadata should be set explicitly in the GSQL model, using annotations. Metadata annotations resemble hashtags (eg. `#ratio`, `#units=usd`) that can be inlined or written above the object they decorate.
+For this reason, some metadata should be set explicitly in the GSQL model, using annotations. Metadata annotations resemble hashtags (eg. `#ratio`, `#units=usd`) that can be inlined or written above the object they decorate. `graphene check` validates metadata keys and values.
```
kevinmarr: we don't mention elsewhere in gsql.md what `check` checks, so prob don't need to add this

docs/references/gsql.md
```diff
@@ -138,18 +138,20 @@ There isn't always a SQL expression that can tip Graphene to the semantic meanin
...
 | Key | Inferrable? | Effect |
 |---|---|---|
 | `#ratio` | no | Value is 0–1; rendered as `value × 100%` (e.g. `0.42` → `42%`) |
 | `#pct` | no | Value is already 0–100; rendered as `value%` (e.g. `42` → `42%`) |
-| `#units=<currency>` | no | Adds currency symbol and compacts to K/M/B. Accepted values: `usd`, `eur`, `gbp`, `cad`, `aud`, `jpy` |
+| `#units=<unit>` | no | Adds currency symbol and compacts to K/M/B for currencies. Accepted values: `usd`, `eur`, `gbp`, `cad`, `aud`, `jpy`, `count`, `seconds` |
 | `#timeGrain=<grain>` | yes (from `date_trunc`, `date_bin`, casts) | Controls time axis label format. Values: `year`, `quarter`, `month`, `week`, `day`, `hour`, `minute`, `second` |
 | `#timeOrdinal=<ordinal>` | yes (from `extract`) | Treats values as ordinal positions on a categorical axis. Values: `hour_of_day`, `day_of_month`, `day_of_year`, `week_of_year`, `month_of_year`, `quarter_of_year`, `dow_0s` (0=Sun), `dow_1s` (1=Sun), `dow_1m` (1=Mon) |
 
+The conventional table and field annotations `#description=<text>`, `#format=<text>`, `#color=<text>`, `#hide`, and `#pii` are also valid. Flag annotations accept either bare usage (for example `#ratio`) or `=true`.
```
kevinmarr: - I thought we deprecated value formatting entirely (`#format` here) and for now are just relying on our metadata-inferred defaults. 
- If `color` indeed works, would be good to mention here exactly what it accepts, eg. hex code, etc instead of just `<text>`
- The intent of the table above was to be comprehensive to all the supported keys, so maybe just add all these to the table instead of having a separate line

docs/references/gsql.md
```diff
@@ -138,17 +138,23 @@ There isn't always a SQL expression that can tip Graphene to the semantic meanin
...
 |---|---|---|
 | `#ratio` | no | Value is 0–1; rendered as `value × 100%` (e.g. `0.42` → `42%`) |
 | `#pct` | no | Value is already 0–100; rendered as `value%` (e.g. `42` → `42%`) |
-| `#units=<currency>` | no | Adds currency symbol and compacts to K/M/B. Accepted values: `usd`, `eur`, `gbp`, `cad`, `aud`, `jpy` |
+| `#currency=<code>` | no | Adds currency symbol and compacts to K/M/B. Accepts ISO 4217 currency codes like `USD`, `EUR`, and `JPY` |
+| `#unit=<unit>` | no | Physical unit label. Any non-empty value is accepted |
 | `#timeGrain=<grain>` | yes (from `date_trunc`, `date_bin`, casts) | Controls time axis label format. Values: `year`, `quarter`, `month`, `week`, `day`, `hour`, `minute`, `second` |
 | `#timeOrdinal=<ordinal>` | yes (from `extract`) | Treats values as ordinal positions on a categorical axis. Values: `hour_of_day`, `day_of_month`, `day_of_year`, `week_of_year`, `month_of_year`, `quarter_of_year`, `dow_0s` (0=Sun), `dow_1s` (1=Sun), `dow_1m` (1=Mon) |
+| `#description=<text>` | no | Description text for a table or field. `--` comments are also collected as descriptions |
+| `#hide` | no | Marks a table or field as hidden. |
```
dylanscott: This does not actually have any effect currently (#23). Maybe we should remove until we implement it?

kevinmarr: yeah i agree - no need to document if it doesn't do anything

docs/references/gsql.md
```diff
@@ -138,17 +138,20 @@ There isn't always a SQL expression that can tip Graphene to the semantic meanin
...
 
 #### Recognized metadata
 
 | Key | Inferrable? | Effect |
 |---|---|---|
 | `#ratio` | no | Value is 0–1; rendered as `value × 100%` (e.g. `0.42` → `42%`) |
 | `#pct` | no | Value is already 0–100; rendered as `value%` (e.g. `42` → `42%`) |
-| `#units=<currency>` | no | Adds currency symbol and compacts to K/M/B. Accepted values: `usd`, `eur`, `gbp`, `cad`, `aud`, `jpy` |
+| `#currency=<code>` | no | Adds currency symbol and compacts to K/M/B. Accepts ISO 4217 currency codes like `USD`, `EUR`, and `JPY` |
+| `#unit=<unit>` | no | Physical unit label. Any non-empty value is accepted |
```
kevinmarr: How about 

>Appends the provided value to the end of labels in visualizations, eg. `unit=minutes` appends ` minutes` (or ` (minutes)` on axes).

(That's how it behaves, right?)

dylanscott: It does now!

kevinmarr (approved):

rogueg (approved):

PR #391, opened 2026-05-09
Extend timeOrdinal axis styling to all timePart extracts

kevinmarr:
I discovered that if you mapped a `extract(year ...)` to an x-axis, the splitLines/axisLines config was not matching our design, similar to what @rogueg just fixed for time ordinals. 

Turns out that year isn't really an ordinal per se, it's just a time part, which is why Grant's fix didn't cover it. There are also more ordinal-less time parts than just year, so this fix ought to cover those, too.

This PR hides splitLines/axisLine and uses an integer formatter on value axes backed by any temporal extract (`extract(year ...)`, decade, century, ...), matching what PR #373 did for `timeOrdinal`. Also widens the paired-y-axis-line guard so the same axes drop the redundant left edge line.


rogueg:
Closing, as we should just get rid of timePart. It's redundant with ordinal. Will open a different PR

PR #394, opened 2026-05-10
disable echart's smoothing

dataders:
maybe this is anomalous for my data (and opinion), but i found the ECharts smoothing rather jarring and not to my taste.

image


kevinmarr:
Not the first time I've heard this, so I'm inclined to acquiesce :) 

However we don't yet have a contributors license agreement so I'm going to do this in my own PR, and perhaps add a property to let this be toggle-able.

PR #396, opened 2026-05-11
add `graphene run --headless` flag

dylanscott:
This PR updates `graphene run` to add a `--headless` flag toscreenshot dashboards using a headless browser via Playwright. I think this approach has a few advantages

- It is transparent to the user, and addresses Kevin's [comment](https://github.com/graphene-data/graphene/pull/383#issuecomment-4408015891) about the visible tab opening being disruptive
- I have confirmed it allows `graphene run` to work in Cursor's sandbox environment
- IMO it is a simpler implementation, and avoids WebSocket screenshot listener and forked `html2canvas`

The main drawback is that it means the CLI has to pull down a browser build - but that part is honestly not as bad as I expected. It uses the Chromium Headless Shell build whose download zip is ~80-100MB depending on platform. Though there is some added complexity around managing this install, including a `graphene install-browser` command that can be used if the postinstall script didn't install it for you.


rogueg:
I rather like that it uses my browser and opens a tab, but it's possible I'll be an outlier here.

Not being able to use `open` in cursor is def a problem, I wonder when they changed their sandbox to prevent this. Cursor has a built-in browser, and I wonder if we should be trying to use it. There's some ux advantage, like having it side-by-side with code, and the ability to reference elements. Thing is idk if we can do this from a CLI, and we might just have to ask the agent to "navigate your browser too localhost:3000/my_page, then `graphene run my_page.md`". Feels unsettlingly non-deterministic, but I bet it'd work.

I don't mind the 80-100mb download, but I would want to make sure that downloading chrome wouldn't hang other agents in other ways. It seems possible that the sandbox would prevent an agent from running `install-browser`, and modern package managers block post-install by default.

The other thing to consider is the browser-level cache. If we have two browsers (playwright for the agent, and one for the user) you're doing twice as many queries against the db.

kevinmarr (approved):

PR #401, opened 2026-05-12
CLI update notifier

dylanscott:
This PR adds a first pass at update notifications for the Graphene CLI. I did some research on best practices here and went with an approach similar to the [update-notifier](https://www.npmjs.com/package/update-notifier) package. To avoid slowing down CLI commands, the CLI periodically (no more than once a day) checks for new versions and stores the info in a cache file in `~/.config/graphene`. We check the cache data when running CLI commands and if we haven't notified about that version in the last week we notify (with a link to the changelog/GH release).

### DIFF
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -16,6 +16,7 @@ import {getConnection, runQuery} from './connections/index.ts'
 import {printDiagnostics, printTable} from './printer.ts'
 import {listMdFileQueries, runMdFile, runNamedQueryFromMd} from './run.ts'
 import {CliTelemetry, getPresentFlags, getWorkspaceScanCounts, type TelemetryCommand} from './telemetry/index.ts'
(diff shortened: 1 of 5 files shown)


kevinmarr:
lgtm - qq though. does the `create` tool create a gitignore file? if so, should it add a line to gitignore this new .config folder?

dylanscott:
That `.config` folder is in the home directory, which we use for caching credentials from `graphene login`. Though now that I say that, that will probably cause issues with agent sandboxes since it's outside of the workspace. I will move it into the same `node_modules/.graphene` folder we are using for other stuff, at the cost of it potentially getting blown away if a user clears node_modules

kevinmarr:
Yeah that sounds good, blowing it away is not a big consequence.

kevinmarr (approved):

PR #405, opened 2026-05-13
Apply title case to some snake-cased stuff

kevinmarr:
Scatter axis labels and wide-form legends were displaying raw snake_case field names. Apply formatTitle in the enrichment pipeline so all wrapper-supplied axis names and series names render in Title Case regardless of chart type. Runs before splitBy expansion so generated series names (raw data values) pass through untouched.

https://claude.ai/code/session_01RRPL5xwU4crHKEH37SXG4P


ui/component-utilities/enrich.ts
```diff
@@ -241,6 +245,18 @@ function buildSplitSeries(template: SeriesWithGroupingHint, datasetId: string, n
...
 
+// Convert snake_case field names (used as axis/series labels) into Title Case so
+// chart labels stay consistent regardless of how the underlying column is named.
+// Runs before splitBy expansion so generated series names (raw data values) are left alone.
+function titleCaseFieldLabels(config: NormalConfig) {
+  for (let axis of [...config.xAxis, ...config.yAxis]) {
+    if (typeof axis?.name === 'string' && axis.name.length > 0) axis.name = formatTitle(axis.name)
+  }
+  for (let series of config.series) {
+    if (typeof series?.name === 'string' && series.name.length > 0) series.name = formatTitle(series.name)
```
rogueg: Changing the series name might be problematic, as I believe we use this to key into the dataset in some cases. Prob worth a test to cover that case.

kevinmarr: Good flag. Took a look — `series.name` is read in exactly two places in `enrich.ts` (line 240 only writes; line 613 reads it as a legend-selection key inside `stackedBarCornerRadius`). Everything else keys off the dataset id and `series.encode`, which we're not touching. The legend-selection read is consistent because ECharts emits its `legendselectchanged` event with the post-enrichment `series.name`, so the persisted selection always lines up with the title-cased name on the next render.

ECharts' own data binding uses `series.encode` when it's present (which all our wrappers set explicitly), and the base dataset has `dimensions` defined, so `series.name` isn't a fallback binding key in our pipeline.

Added `ui/tests/enrich.test.ts` to lock this in — covers wide-form axis/series formatting + `encode` preservation, splitBy data values not being mutated, and legend-selection round-tripping across renders.

---
_Generated by [Claude Code](https://claude.ai/code)_

rogueg:
Yeah, I think claude is confidently lying to you. Other parts of the config (like the legend) can refer to series by name. Rewriting will break those charts.

What is safe, and I just added, is rewriting the series names within our thin wrappers. This works because the wrappers don't refer to series by name.

For custom <ECharts> in theory it might be possible to make this work, but it'd take a lot more complexity than I think it's worth, and anyway it's kind of against the spirit of this component which aims to only set defaults, and leaves what you explicitly set untouched.

There's also an interesting question of if we should format series when `splitBy` is used on tall data. I opted to not do this, on the thinking that for wide data, there's nothing stopping you from having data formatted the way you want in the db, and in fact us reformatting might lead to undesirable results that you can't override.

rogueg (approved):

kevinmarr:
Yay thanks!!

PR #418, opened 2026-05-19
Dependabot upgrades

rogueg:
https://github.com/graphene-data/graphene/security/dependabot/130

This only affects `@vscode/test-cli`, and is not a vulnerability we need to worry about, but I wanted to use it to test our process.


rogueg:
I'm bypassing to deploy this to test out our emergency process.

kevinmarr (approved):

PR #419, opened 2026-05-20
warehouse-backed query caching

dylanscott:
This PR adds a first pass at query caching on the backend, backed by supporting warehouses. This scheme does not store any query result data at all, just a metadata mapping used to utilize the warehouse's own result cache:

- Snowflake and BigQuery each supporting pulling the results of any queries in the last 24 hours, with different APIs
  - Snowflake exposes a queryId and allows that query's results to be queried with SQL
  - BigQuery allows previous jobs' results to be pulled via the API
- ClickHouse exposes a managed version of this, where we opt-in to using the query cache (and potentially stale results), and we configure a 24hr TTL, but ClickHouse handles pulling from the cache for us when we run the query. Cache metadata is returned in headers

I wired the cache metadata up to a placeholder-level UI, which looks like this:

Screenshot 2026-05-20 at 10 40 50 AM

We may want to do a pass on making this look nicer before merging. Additionally, this change interacts with the existing browser cache layer. I opted to make the UI treat results from browser vs backend/warehouse cache the same. Whether served from the browser cache or backend cache we will show the age of the results with the option to refresh.

### DIFF
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -105,7 +105,7 @@ program
       let analysis = analyzeWorkspace({config, files: files.filter(file => file.path != 'input').concat({path: 'input', contents: gsql})}, 'input')
       let [query] = validateInputQuery(analysis, exit)
       let sql = renderSql(query, inputs, exit)
-      let res = await runQuery(sql, {cache: true})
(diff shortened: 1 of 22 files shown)


dylanscott:
going to attack this from a different direction

PR #427, opened 2026-05-22
tweak `QueryConnection` API

dylanscott:
This PR is a preparatory refactor of the `QueryConnection` API to support a caching layer via the cloud query proxy


cli/auth.ts
```diff
@@ -6,11 +6,17 @@ import path from 'path'
...
+const LIVE_AUTH_CLIENT_ID = 'connected-app-live-8264d0af-df18-4021-af96-157482d17856'
 
 export const AUTH_SCOPES = 'offline_access'
 
+export function authClientId() {
+  if (process.env.AUTH_CLIENT_ID) return process.env.AUTH_CLIENT_ID
+  if (process.env.STYTCH_PROJECT_ID?.startsWith('project-test-')) return TEST_AUTH_CLIENT_ID
+  return process.env.NODE_ENV == 'test' ? TEST_AUTH_CLIENT_ID : LIVE_AUTH_CLIENT_ID
+}
+
```
rogueg, line 19: What's this change for?

dylanscott: Without it, `AUTH_CLIENT_ID` would get initialized at module load time, which worked out to before the `loadConfig` in `cli.ts` had loaded the `.env` files. The alternative would be ensuring we lazily import `cli/auth.ts` which felt brittle, but this level of indirection is enough to pick up values from `.env`

lang/index.d.ts
```diff
@@ -6,6 +6,14 @@ export interface QueryResult {
...
   hash?: string // hash of the compiled sql for caching
   sql?: string
+  cache?: QueryCacheMetadata
+}
+
+export type QueryCacheMetadata = {
+  provider: 'bigquery' | 'snowflake' | 'clickhouse'
+  status: 'hit' | 'miss' | 'refresh' | 'native'
+  createdAt?: number
+  expiresAt?: number
```
rogueg, line 16: small nit, but should we just put `hash` on here, so all the cache info is colacated?

I'd also be a fan of simplifying and just putting `runAt` or somesuch directly on QueryResult. It's interesting for reasons beyond just the cache, and I don't think we need provider/status/expiresAt client side, do we?

cli/auth.ts
rogueg: Ah gotcha. lgtm

Change for a different PR, but I wonder if we should adopt a pattern of importing a `loadEnv` script at the top of cli (and server.ts for that matter). It's generally simpler to reason about stuff if you assume the env never changes.

dylanscott:
I'm going to unstack/just include these changes in #431. I was trying to isolate just the changes needed for the cloud side but I don't think it's buying any clarity at this point

PR #431, opened 2026-05-25
cache staleness indicator

dylanscott:
This PR adds a cache staleness indicator showing the age of the oldest query on the page with an option to refresh the queries. Staleness is considered both for browser cached entries (which requires us to store a little more metadata), as well as the new warehouse caching layer.

The indicator UI is now more minimal and in the top-right margin:

Screenshot 2026-05-27 at 10 49 00 AM


kevinmarr:
I think it was to remove the pop tip and just have the text render directly on the page. I think we could minimize the text significantly ie "28m ago [refresh icon]"

rogueg:
Yeah, agree about the viz. I'd also prob put it in the top right margin, even with the nav icon.

ui/internal/queryEngine.ts
```diff
@@ -120,22 +145,32 @@ async function fetchWithCache(req: QueryRequest): Promise<QueryResult> {
     throw err
   }
 
-  cacheWrite(hash, response.clone())
```
rogueg, line 123: The lack of an `await` here was intentional iirc, but def worth documenting.

ui/internal/LocalApp.svelte
```diff
@@ -74,13 +79,37 @@
...
+    return remainingMinutes ? `${hours}h ${remainingMinutes}m ago` : `${hours}h ago`
+  }
 </script>
 
 <SidebarToggle style='position:fixed;top:2rem;left:2rem;opacity:0.3;' />
 <Sidebar>
   <PageNavGroup files={navData} />
 </Sidebar>
 
+{#if cacheAge}
```
rogueg: Should this (and the formatting logic) all be a component?

ui/internal/queryEngine.ts
```diff
@@ -27,14 +29,27 @@ export interface QueryRequest {
...
+  browserCache?: BrowserCacheMetadata
+}
+
+export type QueryFetcher = (req: QueryRequest, options?: {refresh?: boolean}) => Promise<QueryFetchResult>
 
 let runPending: Promise<void> | null = null
+let refreshNextRun = false
 let queries = [] as QueryNode[]
 let queryResults = {} as Record<string, {rows: any[]; fields?: Field[]}>
+let cachedQueryTimestamps = new Map<QueryNode, number>()
```
rogueg: I'd just add this on to queryNode

cli/connections/index.ts
```diff
@@ -166,11 +166,18 @@ async function importConnection<T>(load: () => Promise<T>, packageName: string,
   }
 }
 
-export async function runQuery(sql: string, options?: QueryOptions): Promise<QueryResult> {
+interface ProxyQueryOptions {
+  cacheControl?: string
+}
+
+export async function runQuery(sql: string, options?: QueryOptions, proxyOptions: ProxyQueryOptions = {}): Promise<QueryResult> {
```
rogueg: Should we just simplify to a single `options` object?

rogueg:
I'm having a hard time tracing the logic in this one. Are there any docs/comments I can read?

For the browser cache, all we have to do is clear out the cache entries used by this page, right?

Then for cloud, we also need to set `cache-control: no-cache` on query requests?

The cloud one is interesting, because depending on your use case you might not need cache-control, just clearing the browser cache and refreshing the page would be sufficient.

ui/internal/clientCache.ts
```diff
@@ -4,42 +4,108 @@
...
+  let now = Date.now()
+  let body: QueryResultCacheMetadata = await response
+    .clone()
+    .json()
+    .catch(() => ({}))
+  let createdAt = Number(body?.cache?.createdAt || now)
+  let browserExpiresAt = now + TTL_MS
+  let serverExpiresAt = Number(body?.cache?.expiresAt || browserExpiresAt)
+  let expiresAt = Math.min(browserExpiresAt, serverExpiresAt)
+  await store.put(cacheUrl(hash, createdAt, expiresAt), response)
```
rogueg: I haven't thought this through, but my gut is that the browser should be unaware of the server cache state. What's the thinking here?

dylanscott:
Took another pass incorporating your feedback - hopefully it's clearer now, let me know what you think

kevinmarr:
Looks good. Minor nit is that I think the top row is supposed to be center-aligned, not top-aligned:

Screenshot 2026-05-27 at 10 40 08 AM

dylanscott:
@kevinmarr this look better?

Screenshot 2026-05-27 at 10 49 00 AM

kevinmarr:
Yeah! 👍

cli/connections/index.ts
```diff
@@ -181,7 +190,7 @@ export async function runQuery(sql: string, options?: QueryOptions): Promise<Que
 
   let conn = await getConnection()
   try {
-    return await conn.runQuery(sql, options)
+    return await conn.runQuery(sql, params === undefined ? undefined : {params})
```
rogueg: Is this stacked on your other PR? I thought you refactored to have a single QueryOptions, but I might be misremembering

ui/internal/clientCache.ts
rogueg, line 1: I don't totally grok the logic change in this file, maybe some comments to make it easier to read?

Or is there no logic change here? Is it just moving code around?

cli/connections/index.ts
dylanscott: yes, #427 refactors to make the 2nd `runQuery` arg a `QueryOptions` object rather than just the params, and this path extends so you can specify `cacheControl`.

I don't know why the agent chose to do the ternary here. I believe unconditionally passing `{params}` would be the same (it's not a required field on `QueryOptions`)

ui/internal/QueryCacheStatus.svelte
```diff
@@ -0,0 +1,86 @@
...
+  function formatCacheAge(createdAt: number | undefined, currentTime: number) {
+    if (!createdAt) return ''
+
+    let minutes = Math.max(0, Math.floor((currentTime - createdAt) / 60_000))
+    if (minutes < 1) return ''
+    if (minutes < 60) return `${minutes}m ago`
+
+    let hours = Math.floor(minutes / 60)
+    let remainingMinutes = minutes % 60
+    return remainingMinutes ? `${hours}h ${remainingMinutes}m ago` : `${hours}h ago`
```
rogueg, line 32: prob worth supporting days too? Dno if we need to go beyond that.

ui/internal/queryEngine.ts
```diff
@@ -27,14 +30,26 @@ export interface QueryRequest {
...
 
-export type QueryFetcher = (req: QueryRequest) => Promise<QueryResult>
+export interface PageCacheState {
+  oldestCreatedAt?: number
+  loading: boolean
+}
+
+interface QueryFetchResult {
+  result: QueryResult
+  browserCache?: BrowserCacheMetadata
```
rogueg: I'd just put `runAt` on `QueryResult` and then you can drop this type. It's kinda confusing to have both QueryResult and QueryFetchResult, I can already see myself never remembering which is which.

ui/internal/clientCache.ts
dylanscott: The main change is that we need to know when a browser cache entry was created in order to show its age, so instead of storing `expires` in the cache URL we store `createdAt` and compute the expiry with the TTL

ui/internal/QueryCacheStatus.svelte
dylanscott: FWIW we can't support a longer than 24h TTL on the cloud/backend cache (that is as long as Snowflake and BigQuery keep query results around), so you can only end up with over a day by pulling a soon-to-expire entry into the browser cache for 2h (or with an old tab I guess)

ui/internal/queryEngine.ts
```diff
@@ -27,14 +30,26 @@ export interface QueryRequest {
...
+export type QueryFetcher = (req: QueryRequest, options?: {refresh?: boolean}) => Promise<QueryFetchResult>
 
 let runPending: Promise<void> | null = null
+let refreshNextRun = false
 let queries = [] as QueryNode[]
 let queryResults = {} as Record<string, {rows: any[]; fields?: Field[]}>
 
 let queryFetcher: QueryFetcher = fetchWithCache
 export const setQueryFetcher = f => (queryFetcher = f)
+export const pageCacheState = writable<PageCacheState>({loading: false})
```
rogueg: Nice! I like this. I wonder if we should call it `queryState`, since there is non-cache stuff like the loading state, but that's a nit.

ui/internal/queryEngine.ts
```diff
@@ -86,31 +104,37 @@ async function runNode(n: QueryNode) {
...
   })
   let hash = response.headers.get('ETag') || ''
 
   // cache hit. Read data out of the browser cache and return it
   if (response.status == 304) {
-    return await cacheRead(hash)
+    let cached = await cacheRead<QueryResult>(hash).catch(() => null)
+    if (cached) return {result: cached.result, browserCache: cached.cache}
+    // A missing browser-cache entry means our advertised hash list was stale; retry without that hash.
+    if (allowBrowserCacheRetry) return fetchWithCache({...req, hashes: hash ? req.hashes.filter(h => h != hash) : []}, options, false)
```
rogueg: Can this ever happen? IIRC, the original idea was that we only send hashes of things we have, so the read should never fail.

ui/internal/queryEngine.ts
```diff
@@ -120,22 +144,33 @@ async function fetchWithCache(req: QueryRequest): Promise<QueryResult> {
     throw err
   }
 
-  cacheWrite(hash, response.clone())
-  return await response.json()
+  // Cache writes are best-effort and should not block rendering fresh query results.
+  void cacheWrite(hash, response.clone())
+  return {result: await response.json()}
```
rogueg: Thanks for fixing up my bad comment hygiene!

ui/internal/clientCache.ts
rogueg: But don't we already know that, because it's in the QueryResult (and IIRC we then put it on the QueryNode)?

ui/internal/QueryCacheStatus.svelte
rogueg: Yeah, I'm thinking about the case where you leave a tab open for a long time, and then we show it as `5356m ago`

ui/internal/clientCache.ts
dylanscott: I suppose that's true. The way this was written before seemed agnostic to the type of the value, which is why I assume it was storing expiry in the URL. I will refactor to tie it explicitly to the `QueryResult` type and pull `createdAt` from there

rogueg: The original intent of putting the expiration in the url was so that when getting the list of hashes to send, we could automatically omit ones that we know are stale without having to read their bodies.

For showing in the ui, yeah, I think the QueryResult/QueryNode stuff you have would do it just fine.

There's an interesting question here of "should the browser cache expiry be tied to the server cache in some way, instead of just a flat 2h?" Seems like maybe it should: a query on the server is def valid for more than 2h, and potentially longer than 24 if we ever do our own caching.

rogueg: Maybe we should remove the expire/createdAt in the cache url entirely. That seems more correct (and is easier to reason about), but then the cache would grow unbounded. Maybe for now we just bump it to 24h, and leave the logic as it is?

ui/internal/queryEngine.ts
dylanscott: I think you're right - I got confused because I took the `any` out of the return signature so now it requires a non-null assertion, but it makes sense that if we sent the hash we'll be able to read it

dylanscott:
@rogueg I think all feedback addressed, and I opted just include the changes from #427 in this as one PR rather than split

rogueg (approved):
This looks great!

PR #438, opened 2026-05-26
Don't coerce input strings to dates

rogueg:
String inputs are always strings. Not sure why we added this, but it feels like the wrong solution.

---


dylanscott (approved):
makes sense to me - I think warehouses will typically allow this coercion implicitly with string inputs

PR #446, opened 2026-05-27
proposal: normalize `graphene schema` type output

dylanscott:
This was another pain point the agent encountered a couple of times recently: `graphene schema` would preserve most warehouse-native type spellings (e.g. `varchar` over `string`) in its output, though there were already exceptions (array syntax). The agent (understandably) seems to want to bootstrap its GSQL models by copying the `graphene schema` output over as a starting point and adding any additional modeling on top. When the native types are used in the model this breaks, and particularly with some of ClickHouse's more involved type syntax (e.g. `Enum8('CSH' = 1, 'CRE' = 2)`) the models would fail to parse with very cryptic errors (like "TypeError: Cannot read properties of null (reading 'node')").

This PR proposes to normalize all types in the `graphene schema` output so that it can more easily be used as a starting point for modeling.


rogueg:
Interesting! What are you proposing we normalize to?

Internally we use simplified types, but I wonder if eventually we'll need the underlying database types for smarter type-checking or automatic coercion.

I could also see the argument that an agent might need to know the warehouse type in order to write the correct sql.

kevinmarr (approved):

kevinmarr:
Looks good to me. We never really talked about our data types strategy - eg. do we try to support every underlying data type from the DB or have our own canonical types. I'm not sure I have an opinion

dylanscott:
I am proposing we normalize to the types accepted by GSQL, essentially the `ScalarField` union, since we already normalize array types:

https://github.com/graphene-data/graphene/blob/fc8ead8d653c38d29dde1bb7d6ba6e387af24949/lang/index.d.ts#L33

We do accept some aliases for these in GSQL, which I think has made this less of an issue historically ([SCALAR_TYPE_ALIASES](https://github.com/graphene-data/graphene/blob/fc8ead8d653c38d29dde1bb7d6ba6e387af24949/lang/types.ts#L53)), but I think it would be difficult to support the union of all type syntaxes

rogueg:
Yeah, I guess the question is "are there things we/agents cannot easily do if we don't know the actual db type"

dylanscott:
If there are, it feels like we're already losing that information with the normalization we end up doing in our type system (e.g. resolving all timestamp types into a single one).

My position is that `graphene schema` output should be valid GSQL input. That seems to be what the agent expects. That could also be addressed by expanding the GSQL type system/syntax rather than narrowing `graphene schema`, but it's not obvious to me how to do that in a universal way

rogueg:
Oh, I don't think it affects anything today, I'm wondering what future things become harder if we only have simplified types in models. A few possible examples an agent came up with:

* trying to call a function that only accepts integers with a float
* time vs timestamp are collapsed to the same type, but `date_trunc('day', start_time)` would not work if start_time was a `TIME`
* timestamp tz and ntz are lost, so we can't error/warn when you compare them

Using the proper database types is a bit more verbose, but it also seems more future-proof.

PR #448, opened 2026-05-28
support row prop for `<BigValue/>`

dylanscott:
Quick one from the [dogfood run](https://graphenedata.slack.com/archives/C09MZ1LKB7T/p1779973598188339) this morning: The agent was confused that `<Value/>` supports a row offset but `<BigValue/>` didn't.


docs/references/big-value.md
```diff
@@ -17,4 +17,4 @@ Here's an example:
 | data | GSQL query or table name | true | query name | - |
 | value | Column or expression to pull the main value from | true | column name, stored expression name, GSQL expression | - |
 | title | Title displayed above the value | false | string | - |
-| subtitle | Subtitle displayed below the title | false | string | - |
```
dylanscott, line 20: this was documented here but no longer supported

rogueg (approved):

PR #451, opened 2026-06-02
snowflake auth

rogueg:


cli/connections/bigQuery.ts
```diff
@@ -1,4 +1,5 @@
 import {BigQuery, BigQueryDate, BigQueryTimestamp, type BigQueryOptions} from '@google-cloud/bigquery'
+import {readFileSync} from 'fs'
```
dylanscott: nit: Can we make `localDbOptions` async and use `fs/promises`? The postgres connector looks like it does, so I would expect all the callsites already to be async

dylanscott (approved):
one nit re: cutting down on `readFileSync` usage but this `localDbOptions` pattern feels cleaner 👍

PR #462, opened 2026-06-08
CSV export

### LINKED ISSUE #461: Ability to export data for agents and humans
- Add a download icon at the top right of a chart that downloads the chart data as a .csv. Should work both locally and in cloud.
- Add a way to do this via CLI too, for agents. Perhaps as a flag on `run -q`, `run -c`, and `run [gsql]`. This is helpful if the agent needs to send the .csv somewhere, like as an attachment in Slack or email.

Would be nice to support .xlsx as well, but just .csv to start is fine.

dylanscott:
This PR adds support for exporting query results in CSV format. This is exposed in the CLI with a `--format csv` param on `graphene run`, which works when running a raw GSQL query, targeting a query with `-q`, or a chart with `-c`. Exporting the query results backing a chart is also supported in the UI with a download button that appears in the top right on hover, and looks like:

Screenshot 2026-06-09 at 7 39 50 AM

Fixes #461

### DIFF
--- a/cli/cli.test.ts
+++ b/cli/cli.test.ts
@@ -154,6 +154,12 @@ describe('cli run', () => {
     expect(output).not.toContain('at file://')
   })
 
+  it('prints csv for an inline query with --format csv', async () => {
+    let res = await runCli(['run', "select 'a,b' as name, 2 as total", '--format', 'csv'], {cwd: flightDir})
+    expectCliSuccess(res, 'run query as csv')
+    expect(res.stdout).toBe('name,total\n"a,b",2\n')
+  })
+
   it('accepts --port on run commands', async () => {
     let res = await runCli(['run', 'from flights select count() as total', '--port', '4164'], {cwd: flightDir})
     expectCliSuccess(res, 'run query with port')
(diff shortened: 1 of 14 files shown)


kevinmarr:
Epic. My only feedback would be to remove the border around the icon, and the white button background, so it's less "button-y" and just a clickable icon.

dylanscott:
@kevinmarr done

kevinmarr (approved):

rogueg:
Not a thing introduced in this PR, but it's kinda strange we have both runNamedQueryFromMd and runMdFile. Seems like there should only be one path? Maybe runMdFile should just return the data as well as the screenshot? Could clean up later though.

I'd keep the download stuff in its own svelte component if possible, to avoid cluttering ECharts.svelte

Defer to kevin's design expertise, but the button still feels quite cluttery to me. Maybe make it fainter, or only appear on the chart hover?

rogueg:
oh jk it is on hover, nvm!

dylanscott:
Factored out a `<CsvDownload/>` component. The core logic for `runMdFile` vs `runNamedQueryFromMd` are fairly different, but I factored out a `analyzeMdFile` helper to use from both those functions as well as `listMdFileQueries`

PR #467, opened 2026-06-11
don't show zero state while loading

### LINKED ISSUE #466: BigQuery: queries slower than ~10s silently render "Dataset is empty" while the job completes in the background
The below is from Claude, apologies if it has inaccuracies. But the basic issue I face is that queries that do complete, show up in a page with "Dataset is empty - query ran successfully, but no data was returned from the database". Investigating, the only thing I could tell was that these are a bit slower queries.

> **Title:** BigQuery: queries slower than ~10s silently render "Dataset is empty" while the job completes (and bills) in the background
>
> **Summary:** On BigQuery, any page query whose job takes longer than ~10 seconds renders "Dataset is empty - query ran successfully, but no data was returned from the database" — no error is surfaced, while the BigQuery job continues to completion and is billed in full. Dashboards with several heavy queries are hit hardest: concurrent execution pushes each job past 10 s, so most charts come up empty even though every job succeeds.
>
> **Environment:** `@graphenedata/cli` 0.0.20 (bundled `@google-cloud/bigquery` 8.3.1), Node 24, dialect `bigquery`. Reproduced on `main`: `cli/connections/bigQuery.ts:32`.
>
> **Root cause:** (1) `maxResults` → `@google-cloud/paginator` sets `autoPaginate = false` → single API call, no retry/poll loop. (2) No `timeoutMs` → API default ~10 s wait → `jobComplete: false` with no rows. (3) `@google-cloud/bigquery` (`src/job.ts`) only errors on incomplete results when `timeoutMs` was explicitly passed; otherwise the callback receives `(null, [])` and `runQuery` resolves with `rows = []`, `totalRows = 0`. The server responds `{rows: []}` and `QueryLoad.svelte` renders the empty-dataset notice; the client cache (`ui/internal/clientCache.ts`, 2 h TTL keyed by SQL hash) then pins the empty result.
>
> **Repro (standalone, mirrors `runQuery` line-for-line):**
>
> ```js
> import {BigQuery} from '@google-cloud/bigquery'
> const client = new BigQuery({projectId: '<project>'})
>
> // ~0 bytes billed, pure compute, returns exactly 1 row, takes >10s
> const slowSql = `
>   with t as (select x from unnest(generate_array(1, 60000)) x)
>   select count(*) as n from t a cross join t b where mod(a.x * b.x, 7919) = 0`
>
> const [job] = await client.createQueryJob({query: slowSql, useLegacySql: false})
> const [rows] = await job.getQueryResults({maxResults: 10000})
> console.log(rows.length)            // 0  (returned at ~10.1s, job still RUNNING)
> // ...wait for job DONE, then:
> const [rows2] = await job.getQueryResults({maxResults: 10000})
> console.log(rows2.length, rows2[0]) // 1  {n: 839951}
> ```
>
> Observed: adapter path returns `rows.length=0, totalRows=0` after 10.1 s with the job state still `RUNNING`; the job finishes at ~181 s and the same call then returns the row. In a real page: fast charts render, slow ones show "Dataset is empty", and BigQuery job history shows every job `DONE` with no errors.
>
> **Suggested fix:** wait for job completion before fetching results:
>
> ```ts
> let [job] = await this.client.createQueryJob({query: sql, useLegacySql: false, params})
> await new Promise((resolve, reject) => {
>   job.on('complete', resolve)
>   job.on('error', reject)
> })
> let [rows] = await job.getQueryResults({maxResults: 10000})
> ```
>
> (or loop on `getQueryResults` until `jobComplete: true`). An in-flight job should never surface as a successful empty result — especially since the warehouse bills the query regardless.

dylanscott:
Fixes #466

### DIFF
--- a/ui/tests/value.test.ts
+++ b/ui/tests/value.test.ts
@@ -28,6 +28,51 @@ describe('<Value/>', () => {
     await expect(sharedPage.getByText('—')).toBeVisible()
   })
 
+  test('renders loading state while query is pending', async ({sharedPage, server}) => {
+    await sharedPage.goto(`${server.url()}/__ct`)
+    await sharedPage.evaluate(async () => {
+      let g = window.$GRAPHENE
+      let originalQuery = g.query
+      let originalUnsubscribe = g.unsubscribe
+      ;(window as any).__restoreQueryHooks = () => {
+        g.query = originalQuery
+        g.unsubscribe = originalUnsubscribe
+        delete (window as any).__queryCallback
+        delete (window as any).__restoreQueryHooks
+      }
+
+      g.query = (_data: string, _fields: Record<string, string>, callback: (result?: unknown) => void) => {
+        ;(window as any).__queryCallback = callback
+        callback()
+        return 'loading-test'
+      }
+      g.unsubscribe = () => {}
+
+      if (window.__inst) g.svelte.unmount(window.__inst)
+      let container = document.getElementsByTagName('main')[0] || document.createElement('main')
+      if (!container.isConnected) document.body.appendChild(container)
+      container.innerHTML = '<div id="component-test"></div>'
+      let target = document.getElementById('component-test')
+      if (!target) throw new Error('component test target was not created')
+      window.__inst = g.svelte.mount(g.components.Value, {target, props: {data: 'slow_query', column: 'value'}})
+      await new Promise(r => requestAnimationFrame(() => requestAnimationFrame(r)))
+    })
+
+    try {
+      let component = sharedPage.locator('#component-test')
+      await expect(component.getByRole('status')).toBeVisible()
+      await expect(component.getByText('Dataset is empty')).toHaveCount(0)
+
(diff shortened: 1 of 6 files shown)


kevinmarr (approved):

ui/tests/value.test.ts
```diff
@@ -28,6 +28,51 @@ describe('<Value/>', () => {
...
+      await sharedPage.evaluate(async () => {
+        ;(window as any).__queryCallback({rows: [{value: 42}], fields: [{name: 'value', type: 'number'}]})
+        await new Promise(r => requestAnimationFrame(() => requestAnimationFrame(r)))
+      })
+      await expect(component.getByText('42')).toBeVisible()
+    } finally {
+      await sharedPage.evaluate(() => (window as any).__restoreQueryHooks?.())
+    }
+  })
+
```
rogueg, line 74: Ideally we'd just have a screenshot test that asserts the loading state looks like we'd expect. That's my bad for not adding one in the first place!

PR #468, opened 2026-06-11
allow fully custom styling

dylanscott:
This PR allows fully custom styling by supporting `<style>` tags as well as layout HTML like `<div>`, `<span>`, etc. We sanitize CSS to prevent legacy JS execution hooks like `expression`. It also restricts dynamic markup in dashboards, as it turns out we were inadvertently supporting some JS in dashboards due to the way we were passing through Svelte syntax/expressions (e.g. this could be as simple as a`{alert(1)}` block in markdown, or in component props).

As an example, I had an agent generate a custom styled dashboard in the flights example:

2026-06-11T18-30-03-469Z


kevinmarr:
Hype!!! Q: Do we whitelist the domain that serves google fonts? Is that easy or non-trivial to allow?

dylanscott:
We don't do any whitelisting for Google Fonts. If we want to allow it I would be hesitant to try to whitelist e.g. `<link>` tags to Google Fonts specifically. Maybe we could add frontmatter for this?

kevinmarr:
i think a big part of custom styling is the ability to pick fonts, so would be keen to solve this somehow - going to set up time to discuss

dylanscott:
@kevinmarr After thinking about it more I opened it up so you can import from Google Fonts or other providers. This does allow CSS to make external HTTP requests, but given that we don't support dynamism in this context you couldn't use this to exfiltrate data, only to ping a fixed endpoint when a page is loaded, which feels okay

cli/sanitization.ts
```diff
@@ -0,0 +1,322 @@
+export const GLOBAL_HTML_ATTRS = ['class', 'id', 'role', 'aria-*', 'data-*']
```
rogueg, line 4: I'm wondering if we should use something like domPurify to sanitize. vibe-coded sanitization makes me nervous

cli/mdCompile.ts
```diff
@@ -122,6 +130,10 @@ export function injectComponentImports() {
     markup: ({content, filename}: {content: string; filename: string}) => {
       if (!filename.endsWith('.md')) return // only auto-import components for md files
       content = liftInlineEChartsConfig(content)
+      validatePreprocessedMarkup(content)
+      let pageStyles = extractPageStyles(content)
+      content = escapeSvelteTextExpressions(pageStyles.html)
+      if (pageStyles.css.trim()) content = `<svelte:head><style>${pageStyles.css}</style></svelte:head>\n${content}`
```
rogueg, line 136: If you put the styles in <head>, what happens as you navigate between pages? Does it get removed when you unmount?

I would have expected css rules scoped to the Page, that way they wouldn't leak out and affect other parts of the ui

dylanscott: As far as I can tell we do not do client-side navigation. Navigating between pages triggers a full page load, so other pages will not include the styles

cli/sanitization.ts
dylanscott: This list in particular is only used as an argument to [sanitize-html](https://www.npmjs.com/package/sanitize-html). But I took a pass at the other sanitization logic and switched to postcss for CSS sanitization and using the Svelte parser for rejecting dynamic syntax. There's still a bit of low-level tag matching logic, but it's to handle extracting the bodies of `<style>` tags (and reused for extracting inline `<ECharts/>` configs) as we need to combine these into one tag for Svelte

kevinmarr (approved):

PR #470, opened 2026-06-11
motherduck support

dylanscott:
This PR adds support for connecting to MotherDuck, which is pretty straightforward since we already support DuckDB syntax.


kevinmarr:
Do we need to add a section to docs/setup.md?

I think there is also a sentence on /readme.md that lists all the supported databases that we'll want to mention Motherduck in

rogueg (approved):

PR #473, opened 2026-06-15
support screenshotting table with `graphene run -c`

dylanscott:
This PR updates `graphene run -c` to support screenshotting table components as well as charts, as there are situations (e.g. `contentType=colorscale`) where the agent needs to visually iterate on a table.


kevinmarr (approved):
didn't even realize this didn't work before

PR #480, opened 2026-06-17
don't do screenshot snapshots for ecomm example

dylanscott:
~I'm not sure what's going on here but the screenshot snapshots for the ecomm example, added in #476, are consistently failing for me in CI.~ The ecomm dataset is continuously updated, so with how it is currently configured the graphs are expected to drift day-by-day as more data comes in.

### DIFF
--- a/ui/tests/run-examples.test.ts
+++ b/ui/tests/run-examples.test.ts
@@ -79,7 +79,7 @@ let testsDir = path.dirname(fileURLToPath(import.meta.url))
 let coreDir = path.resolve(testsDir, '../..')
 let examplesDir = path.join(coreDir, 'examples')
 
-async function runExample(exampleName: string, page: Page) {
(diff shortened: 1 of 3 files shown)


kevinmarr:
The new line graphs are showing a big drop on the last data point, which looks worse than before - I've seen charts (in general, outside Graphene) do this when they have partial data on the last time grain. Is it possible that we started plotting zeros? Or some kind of time zone shift?

dylanscott:
actually the issue here is that the ecomm example dataset is continuously updated, so it is not a candidate for stable screenshot tests unless we update it to use a stable range of data (and assuming old data is never updated). I don't believe anything has changed about our chart rendering (the previous snapshot was created yesterday). Updated the PR to suppress the screenshot comparison for the ecomm example

kevinmarr:
Aha, that makes sense.

kevinmarr (approved):

PR #481, opened 2026-06-17
expose preprocessMarkdownSvelte for cloud path

dylanscott:
This PR updates the custom styling support from #468 to expose the style handling in our preprocessing via a `preprocessMarkdownSvelte` so that the cloud path can use it. Corresponding cloud fix in graphene-data/co#173


rogueg:
Would just a plain `<style>` tag in the svelte code not work?

dylanscott:
@rogueg I was running into some issues with svelte style scoping interacting with stuff with a normal svelte `<style>`, but I just changed it back to putting it in `<svelte:head>` and it seems to work nicely, just needed the change in graphene-data/co#173 to get the styles to unmount on nav

PR #482, opened 2026-06-17
flesh out duckdb function library

dylanscott:
This PR adds substantially all of the remaining duckdb function library to `duckDbFunctions.ts`, with some exceptions like `json_tree` and `json_each` which are "table valued" and not currently supported with our type system. This required some changes outside of just listing the function catalogs:

- Richer function return types, like `array_element` which corresponds to the element type of an input array type
- Addition of a first-class `map` type


kevinmarr (approved):
noice

PR #483, opened 2026-06-18
render authored content in iframe

dylanscott:
This PR updates our rendering pipeline to render authored pages/dashboards in a sandboxed iframe, and removes most of our sanitization logic. This allows pages to specify `<script>`s, but runs them with a locked down CSP that prevents external network requests. Queries are handled via a postMessage API with the parent frame.


kevinmarr:
does this block requests to google fonts?

dylanscott:
@kevinmarr No, but with the content security policy we set now we do have to explicitly whitelist google fonts:

https://github.com/graphene-data/graphene/blob/d740de417a101d18207631683e7ab6f2060e0afd/cli/serve2.ts#L233-L235

kevinmarr:
do you think we should explain in the agent skill what domains are whitelisted? or will the agent not run into problems there

dylanscott:
yeah I can add that to the docs

cli/serve2.ts
```diff
@@ -222,6 +223,38 @@ async function handlePage(server: ViteDevServer, res: ServerResponse<IncomingMes
...
 
+function handleFramePage(res: ServerResponse<IncomingMessage>) {
+  res.setHeader('Content-Type', 'text/html')
+  res.setHeader(
+    'Content-Security-Policy',
+    [
+      "default-src 'none'",
+      "script-src 'self'",
+      "style-src 'self' 'unsafe-inline' https://fonts.googleapis.com",
+      "img-src 'self' data: blob:",
```
rogueg, line 234: this means we can't load images, which is def more secure, but might feel quite limiting in building useful apps.

@kevinmarr thoughts?

rogueg (changes requested):
Neat! I'm still going through it, but at a high level I'm finding some of the logic a bit hard to follow. It's prob worth having an agent go through to add comments and simplify where possible.

cli/sanitization.ts
rogueg, line 1: Should we just delete this file? Seems like all the references to it are mostly leftover

ui/frame.js
rogueg, line 1: this duplicates stuff from web.js, but I'm wondering if we even need it in web.js any more, since rendering components only ever happens in frames now.

ui/internal/runSocket.ts
rogueg, line 1: For local stuff, maybe we just have the CSP allow connections to localhost? That'd mean we need to make no change here (I think), and might simplify other things as well

ui/internal/queryEngine.ts
```diff
@@ -140,6 +141,28 @@ async function fetchWithCache(req: QueryRequest, options: {refresh?: boolean} =
   return await response.json()
 }
 
+export async function runParentQuery(req: QueryRequest, options: {refresh?: boolean} = {}) {
```
rogueg, line 144: This feels off to me, seems like the queryEngine shouldn't have to be aware of the bridge at all, they're separate concerns.

ui/internal/LocalFrame.svelte
```diff
@@ -0,0 +1,155 @@
...
+    if (!isCurrentPageError(payload.err)) return
+    reportCompileError(payload.err)
+    window.$GRAPHENE.appLoading = false
+    bridge.notify('render.ready')
+  })
+
+  onMount(async () => {
+    try {
+      document.fonts.load("12px 'Source Sans 3'")
+      await document.fonts.ready
```
rogueg, line 45: what's this for?

ui/components/ECharts.svelte
```diff
@@ -129,6 +134,19 @@
...
+    let done = false
+    let complete = () => {
+      if (done) return
+      done = true
+      window.$GRAPHENE?.renderComplete?.(renderId)
+    }
+    let afterPaint = () => requestAnimationFrame(() => requestAnimationFrame(complete))
+    chart?.once?.('finished', afterPaint)
+    window.setTimeout(afterPaint, 1000)
+  }
```
rogueg, line 148: What's this for?

cli/sanitization.ts
dylanscott: I think the functions in here are still used by `mdCompile.ts`, though the filename isn't super descriptive of what remains

ui/internal/LocalFrame.svelte
dylanscott: it was migrated from LocalApp's onMount - I'll restore the comment

https://github.com/graphene-data/graphene/blob/fda0272e88edf268e1307d54987a0a264537faaf/ui/internal/LocalApp.svelte#L54-L60

ui/components/ECharts.svelte
dylanscott: I don't love it and can try alternative approaches, but I was running into lots of test failures/flakes due to `waitForLoad` firing while charts were still rendering with the change. I think this in part is due to the iframe itself causing a chart resize after mount

cli/serve2.ts
kevinmarr: right, rendering local images in the workspace should be supported

PR #487, opened 2026-06-19
fix test-examples flights flake

dylanscott:
The `test-examples` CI check has been flaking a lot in CI, and the cause appears to be non-determinism on the delay_heatmap order. Ran 10x locally with the change without a flake. The ordering issue is slightly subtle: The order doesn't determine the layout within the heatmap, it determines the _draw_ order, and cells have a slight overlap so the draw order can change colors on the edges of cells


rogueg:
Nice find! That's the second time we've seen dashboard instability because of a missing order by. I wonder if we should warn on it? Automatically add order to chart queries?

rogueg (approved):

PR #498, opened 2026-07-08
Allow js/css

rogueg:
This removes the sanitization we had, as we want to allow js/css on pages.

Locally this should be fine. On graphene cloud, we'll sandbox pages in an iframe, and I simplified the api for binding url params to inputs in order to support that.


kevinmarr:
Prob need an update to base.md

rogueg:
> Prob need an update to base.md

Yeah, but I'm not sure if it should be in this PR? I kinda want to play with it more to get a sense of where it works and doesn't before agents start using it more widely.

For example, this allows svelte conditionals and it'd be neat to show and hide bits of the ui depending on if a query returns rows, but at the moment there's no api to use for that. It's not even clear how much we want to encourage that, as it ties graphene md to a particular version of svelte.

kevinmarr:
Ah ok. Fine by me

kevinmarr (approved):

PR #509, opened 2026-07-22
Add config.md to references

rogueg:
It wasn't in there, so agents didn't know how to configure graphene.

Also tweaked the build script to put descriptions into the skill for each reference, and folded some of the input references into base.md as they were pretty tiny.


cli/esbuild.mjs
```diff
@@ -37,6 +37,21 @@ await esbuild({
   plugins: [makeAllPackagesExternalPlugin],
 })
 
+// Keep the generated skill's reference list useful for agents, and fail the build when a new reference lacks a description.
+let referenceDocs = {
+  'config.md': 'Project package.json configuration, including database connections, ignored files, env files, telemetry, and dev server settings.',
+  'echarts.md': 'ECharts component usage for custom charts beyond the built-in Graphene chart components.',
+  'gsql.md': 'GSQL query syntax reference for pages, including selects, filters, joins, aggregations, parameters, and SQL passthrough.',
```
kevinmarr: It's more than just the query syntax in pages - it also covers modeling in .gsql

cli/esbuild.mjs
```diff
@@ -37,6 +37,21 @@ await esbuild({
...
+  'model-gsql.md': 'Semantic model GSQL reference for tables, dimensions, measures, joins, namespaces, and database-specific types.',
```
kevinmarr: This is just best practices for modeling, but does not include GSQL reference - that's in gsql.md

kevinmarr: I'm open to splitting half of the GSQL reference into model-gsql.md if you think that makes more sense. But there are features like metadata that are relevant for both modeling and querying

rogueg:
Updated both. I'm happy to leave the gsql files as is, I just wanted to add descriptions to give the agent a better chance of reading the right ones in the right situation.

kevinmarr (approved):

PR #523, opened 2026-07-30
Better input sizing

rogueg:
This makes inputs a bit wider by default (400 v 200px) but also makes them more responsive so if you have many on a single row they wont overlap


kevinmarr:
What does it look like if there is only 1 input on a row with `layout: dashboard`?

rogueg:
Screenshot 2026-07-30 at 12 42 07 PM

They're wider by default, 400px instead of 200px. I was thinking wider would be better for text inputs and search

rogueg:
The `layout` doesn't really affect inputs, except insofar is it sets a max-width on a Row that has multiple inputs.

kevinmarr:
ok. works for me. different inputs will have different default lengths at some point, once i get to overhauling input styles. but i'm not tackling that soon

kevinmarr (approved):

PR #564, opened 2026-08-12
Only use latin woff2 fonts

rogueg:
Our standalone html files were ~4MB, mostly because of fonts. We have a 4 different fonts, and each was including the full range of characters and file formats.

This reduction takes us down to 1.67MB, only 278k of which is fonts.

We could avoid embedding fonts entirely, or reduce the number we use, but idk that it's worth it.

The next biggest contributor is echarts itself, which is 1.1MB

### DIFF
--- a/ui/app.css
+++ b/ui/app.css
@@ -1,13 +1,18 @@
-@import '@fontsource/inter/400.css';
-@import '@fontsource/inter/600.css';
-@import '@fontsource/inter/700.css';
-@import '@fontsource/source-sans-3/400.css';
-@import '@fontsource/source-sans-3/600.css';
-@import '@fontsource/source-serif-4/400.css';
-@import '@fontsource/source-serif-4/400-italic.css';
-@import '@fontsource/source-serif-4/600.css';
-@import '@fontsource/source-serif-4/600-italic.css';
-@import '@fontsource/jetbrains-mono/400.css';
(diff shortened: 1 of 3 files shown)


ui/app.css
```diff
@@ -1,13 +1,13 @@
...
-@import '@fontsource/source-sans-3/400.css';
-@import '@fontsource/source-sans-3/600.css';
-@import '@fontsource/source-serif-4/400.css';
-@import '@fontsource/source-serif-4/400-italic.css';
-@import '@fontsource/source-serif-4/600.css';
-@import '@fontsource/source-serif-4/600-italic.css';
-@import '@fontsource/jetbrains-mono/400.css';
+@font-face { font-family: 'Inter'; font-style: normal; font-display: swap; font-weight: 400; src: url('@fontsource/inter/files/inter-latin-400-normal.woff2') format('woff2'); }
+@font-face { font-family: 'Inter'; font-style: normal; font-display: swap; font-weight: 600; src: url('@fontsource/inter/files/inter-latin-600-normal.woff2') format('woff2'); }
+@font-face { font-family: 'Inter'; font-style: normal; font-display: swap; font-weight: 700; src: url('@fontsource/inter/files/inter-latin-700-normal.woff2') format('woff2'); }
```
kevinmarr, line 5: No italics on headers? Seems reasonable to support

rogueg: We can add it if you like. I didn't change the set of fonts, we never had inter-italic. It looks like we also don't have it for source-sans-3, do we want that one too?

kevinmarr: might as well add it to source sans, sure

kevinmarr (approved):

PR #589, opened 2026-08-21
Better selectAllByDefault

rogueg:
If a multiselect has a ton of values, selectAllByDefault can mean the url gets enormous, potentially getting blocked by the browser or WAF.

I think that if you use selectAllByDefault, we should invert the logic so the params only show the values you're excluding, and your query would be a `NOT IN ($param)`.

We can even enforce this by analyzing queries using dropdown params and ensuring that they use either IN or NOT IN as appropriate.

This is a breaking change in that queries using a selectAllByDefault will need to be changed, though `graphene check` should catch them. It also means saved urls might behave differently now.

I also fixed the empty-set handling, so selecting all/none won't generate invalid sql like `not in ()`

### DIFF
--- a/docs/base.md
+++ b/docs/base.md
@@ -168,7 +168,7 @@ Unsupported:
 Input values are referenced by their `name` as `$name` in Graphene SQL and sync into the page URL query string, preserving state across reloads and shared links.
 
 ### `<Dropdown>`
-Build options from a query using `data` and `value`; optionally use `label` as the displayed column. Other attributes are `title`, `description`, `defaultValue`, `multiple`, `selectAllByDefault`, `noDefault`, and `disableSelectAll`. `name` is required.
+`name` is required. Build options from a query with `data` and `value`; use `label` to display a different column. Optional attributes include `title`, `description`, `defaultValue`, `multiple`, `selectAllByDefault`, `noDefault`, and `disableSelectAll`.
 
 ````md
 ```sql statuses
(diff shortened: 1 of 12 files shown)


kevinmarr:
If this flips the set logic, does it mean you get an enormous URL when only one item is selected?

rogueg:
> If this flips the set logic, does it mean you get an enormous URL when only one item is selected?

If you have `selectAllByDefault=true`, and then you deselect everything, then yes. If you don't have `selectAllByDefault` then the list contains the things that were selected, as it does today.

In theory we could have it figure out which would be more efficient and flip the query back and forth on demand, but that would only cut it in half at most.

kevinmarr:
I worry that this only flips the problem while also creating a breaking change - what other solutions could we explore?

rogueg:
Fair. What makes this tricky imho:

imagine you have `selectAllByDefault` with 50 items, you unselect 3, then you send that url to a coworker. They look at it a few days later, only now there are 60 items. Should they see the exact same 47 you saw, or should they see all minus the 3 you deselected?

Related to this is what query you write for params. Right now it's something like `where foo in $param` or `where foo not in $param`. If we do the thing I propose where the mode (include vs exclude) is dynamic based on what you've selected in the dropdown, we'll have to rewrite this portion of the query.

I could also see punting on the whole thing - having a ux where you might select 100 items or just a few feels like a sub-par page UX. If you select all you probably want them aggregated, if you select just a few you'd likely want them as series.

kevinmarr:
Maybe we need to separate this into sub-problems, then tackle them individually:
- Navigating to a dashboard that has inputs w/ `selectAllByDefault` immediately redirects from `/path` to `/path?<very-long-params-string>`. If the input has not been changed, maybe we should make the URL stay at `/path` so that if you share by copy/pasting the URL, you get the short URL. The subtle change in behavior here is that the naked link keeps updating the default set as it changes over time, but I think that's desirable more often than it isn't.
- How to share the point-in-time static list of all selected items, vs. sharing "all, whatever 'all' is right now". Perhaps we add a Share button that lets you specify these things?
- Defining filters by inclusion vs. exclusion
- Better UX overall - maybe another opportunity for the "smart filter" idea where it's just an open-ended text box that gets translated via LLM into a SQL predicate. Since it is an entire predicate, you just plug it into queries as `where $param`.

rogueg:
@kevinmarr ready for review

kevinmarr:
Did you change anything? I don't know how to see in the GH UI.

rogueg:
I changed it to allow the the dropdown to decide whether to send an include or exclude list, depending on which would have fewer.

kevinmarr:
Got it, so a page can take an additive query param list or a subtractive query param list simultaneously? Flipping from one mode to the other doesn't somehow invalidate old links right?

rogueg:
yeah, an input can express it's values either way, and it gets encoded into the url

kevinmarr (approved):

PR #592, opened 2026-08-21
Echart axis name and legend improvements

rogueg:


ui/component-utilities/enrich.ts
```diff
@@ -403,8 +403,27 @@ function computeTitleLegendAndGridPadding(config: NormalConfig) {
...
+  let hasTopYAxisName = config.yAxis.some(axis => axis?.show !== false && axis?.name && (axis.nameLocation == null || axis.nameLocation === 'end'))
+  if (legendVisible && hasTopYAxisName) {
+    // ECharts selects the plain/scroll legend implementation before applying theme defaults, so the theme's
+    // scroll type is too late. Its runtime scroll option also has an update animation omitted from the public type.
+    let scrollLegend = legend as typeof legend & {animationDurationUpdate?: number}
+    scrollLegend.type ||= 'scroll'
+    scrollLegend.animationDurationUpdate ??= 0
     grid.top = numericOffset(grid.top, 24)
+
+    // ECharts centers end-positioned names over each axis line. Align them outward in the same direction as
```
kevinmarr: Is it possible to align them inward instead? So the left axis name is left-aligned, right axis name is right-aligned? The reason I ask is because in the screenshots below, "Response time" sticks out to the right, creating dead whitespace.

rogueg: Depends on what you mean by left-aligned! afaict, echarts only supports left aligning relative to the grid (ie the actual plot area). So they'd be inwards, but very far inwards.
Screenshot 2026-08-21 at 12 21 20 PM

I didn't like the look of this, but not a hill I want to die on if you feel strongly.

kevinmarr: oh hm i don't like this either. i think at this rate i would prefer a traditional axis name position where it's rotated 90 degrees and vertically centered on the left/right of the chart - then you don't need this extra row of space below the legend anymore.

rogueg: Sadly, the `gap` you specify is relative to the axis line itself, not the bounding box of the tick labels, which is why this one look so lopsided.

Screenshot 2026-08-21 at 5 10 20 PM

kevinmarr: ok, well, my vote then is to keep it on top, centered over the axis line like it is currently. my issue with the outward-projecting is that it is more likely to create whitespace on the sides, which won't align with other charts if in a grid, or could create uneven whitespace if the axis label names are different lengths.

rogueg: fine by me, though fwiw both things might already be true if the tick labels are different widths.

kevinmarr: right, but should be less severe/frequent

kevinmarr (approved):

PR #599, opened 2026-08-24
Preserve metadata through common function kinds

### LINKED ISSUE #587: Preserve metadata through functions/operations
Right now, `min(onboardYear) as earliest` strips the metadata from `earliest` so we no longer know it's a year.

We need a way to figure out which metadata is safe to preserve in what circumstances. We did this narrowly for [casting](https://github.com/graphene-data/graphene/issues/397) a while back, but doing it for most functions/operators would be useful.

rogueg:
Allow function definitions to classify metadata behavior as idempotent, selection, additive, or counting. Centralize propagation in the metadata module, apply it to common functions across dialects, and retain function metadata through window expressions.

fixes #587


kevinmarr:
fixes https://github.com/graphene-data/graphene/issues/587

kevinmarr (approved):

